PR #5, opened 2021-01-05
File upload

Abhinav-grd:


src/components/FullScreenDropZone.tsx
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const FullScreenDropZone = ({
+  children,
+  noClick,
+  closeModal,
+  showModal,
+}) => {
+  return (
+    <>
```
pushkar8723, line 18: [change]
No need to add an extra fragment if it just has one child. Fragments are supposed to be used only where there are multiple root nodes in JSX.

src/pages/_app.tsx
```diff
@@ -81,69 +83,88 @@ const GlobalStyles = createGlobalStyle`
...
+    clearKeys();
+    clearData();
+    setUploadButtonView(false);
+    localForage.clear();
+    const cache = await caches.delete('thumbs');
+    router.push('/');
+  };
+
+  return (
+    <>
```
pushkar8723, line 136: [Change] Remove fragment

src/pages/gallery/components/CollectionDropZone.tsx
```diff
@@ -0,0 +1,84 @@
...
+  width:33%;
+  border-width: 2px;
+  border-radius: 2px;
+  border-color: ${(props) => getColor(props)};
+  border-style: ${(props) => enableBorder(props)};
+  outline: none;
+  transition: border 0.24s ease-in-out;
+`;
+
+const CollectionDropZone = ({
```
pushkar8723, line 31: [question]
Why do we need two dropzone components(FullScreenDropzone and CollectionDropZone)?

[change & note for future]
Also, avoid using the fat arrow function for creating a component. Instead, use functions.
Fat arrow functions are bound to the lexical scope and we don't use this inside a component so an extra function is getting created unnecessarily.

src/pages/gallery/components/Upload.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const [uploadStage, setUploadStage] = useState<UPLOAD_STAGES>(UPLOAD_STAGES.START);
+    const [fileCounter, setFileCounter] = useState({ current: 0, total: 0 });
+    const [percentComplete, setPercentComplete] = useState(0);
+    const init = () => {
+        setProgressView(false);
+        setUploadStage(UPLOAD_STAGES.START);
+        setFileCounter({ current: 0, total: 0 });
+        setPercentComplete(0);
+    }
+    return (<>
```
pushkar8723, line 17: [change]
Remove fragment.

src/pages/gallery/components/UploadButton.tsx
```diff
@@ -0,0 +1,15 @@
+import CollectionSelector from 'pages/gallery/components/CollectionSelector';
+import React, { useRef } from 'react';
+import { Button } from 'react-bootstrap';
+
+const UploadButton = ({ showModal }) => {
+  return (
+    <>
```
pushkar8723, line 7: [change]
Remove fragment.

src/pages/gallery/components/UploadProgress.tsx
```diff
@@ -0,0 +1,30 @@
...
+            centered
+            backdrop="static"
+        >
+            <Modal.Header>
+                <Modal.Title id='contained-modal-title-vcenter'>
+                    Uploading Files
+        </Modal.Title>
+            </Modal.Header>
+            <Modal.Body>
+                {now == 100 ? (
```
pushkar8723, line 19: [change]
Stick to `===`, to have a strict check, and avoid surprises.

src/services/uploadService.ts
```diff
@@ -0,0 +1,396 @@
...
+        let dateString: string = exifData.DateTimeOriginal;
+        var parts = dateString.split(' ')[0].split(":");
+        var date = new Date(Number(parts[0]), Number(parts[1]) - 1, Number(parts[2]));
+        return date.getTime() * 1000;
+    }
+
+    private getLocation(exifData) {
+
+        if (!exifData.GPSLatitude)
+            return { lat: null, lon: null };
```
pushkar8723, line 364: [question]
@vishnukvmd  Should we also have a check for 0,0 coordinates?

pushkar8723 (changes requested):
Delete `package-lock.json` file. We use yarn and so let us stick to just one package manager.

src/pages/gallery/components/CollectionDropZone.tsx
Abhinav-grd: The FullScreenDropzone is just for trigerring the onDrag event on the whole screen It does nothing else. The collectionDropzone handles all the operation. It could be replaced by just a normal elemnt with onDragOver will do that.....didnt know about the HTML onDragOver event earlier thought it was something that React-dropzone provided.

src/services/uploadService.ts
Abhinav-grd: should I make latitude longitude fallback value could be made 0,0 coordinate?

vishnukvmd: No, that looks like a [legit location](https://www.google.com/maps/place/0%C2%B000'00.0%22N+0%C2%B000'00.0%22E/@12.5735552,-4.8134094,3.49z/data=!4m2!3m1!1s0x0:0x0). Omit both `latitude` and `longitude` attributes.

pushkar8723: I meant that if the location present in the EXIF data is `0,0`, then most probably it's wrong. So, should we revert back to `null` in such cases?

Abhinav-grd: ohk....The exif data either has the correct location or is doesn't have location property at all. I haven't found any example when it was 0,0. None the less will add a check and revert back

vishnukvmd: If the EXIF data suggests that it is `0,0`, let's accept it as is. Who knows where the user has been. 🤷

pushkar8723 (approved):

PR #6, opened 2021-01-14
Collection Create and Add to favorite

Abhinav-grd:


.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+package-lock.json
```
pushkar8723, line 7: we should not ignore `package-lock.json` as it helps maintain consistency between dev and prod systems.

src/pages/gallery/components/Collections.tsx
```diff
@@ -57,6 +57,7 @@ export default function Collections(props: CollectionProps) {
         <Wrapper>
             <Chip active={!selected} onClick={clickHandler()}>All</Chip>
             {collections?.map(item => <Chip
+                key={item.id}
```
pushkar8723, line 60: 👍🏼

src/pages/gallery/components/CreateCollection.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const { acceptedFiles, setProgressView, progressBarProps, refetchData, modalView, closeModal, closeUploadModal } = props;
+    const [albumName, setAlbumName] = useState("");
+
+    const handleChange = (event) => { setAlbumName(event.target.value); }
+
+    useEffect(() => {
+        if (acceptedFiles == null)
+            return;
+        let commonPathPrefix: string = (() => {
+            const A: string[] = acceptedFiles.map(files => files.path);
```
pushkar8723, line 19: We should avoid 1 char variable names.

src/utils/common/apiUtil.ts
```diff
@@ -1,3 +1,5 @@
 export const getEndpoint = () => {
-    return process.env.NEXT_PUBLIC_ENTE_ENDPOINT || "https://api.ente.io";
+    const endPoint = process.env.NEXT_PUBLIC_ENTE_ENDPOINT ?? "https://api.ente.io";
+    console.log(endPoint);
```
pushkar8723, line 3: Avoid  `console.log` these will reflect in production logs as well and are also not a very efficient way of debugging. 
I will add a lint check, which will get rid of this.

.gitignore
Abhinav-grd: image

you told to do this na?

pushkar8723: i meant just delete the file. Not add it in git ignore.

pushkar8723 (approved):

PR #9, opened 2021-02-03
File loading single update

Abhinav-grd:


package.json
```diff
@@ -24,6 +24,7 @@
     "react-dom": "16.13.1",
     "react-dropzone": "^11.2.4",
     "react-photoswipe": "^1.3.0",
+    "react-top-loading-bar": "^2.0.1",
```
pushkar8723, line 27: Do we really need to add a dependency for a simple loading bar? Few lines of CSS can take care of it.
Every dependency we add gets shipped to the client. We should be very conservative with what we add.

PR #10, opened 2021-02-03
added InfiniteLoader wrapper for prefetching

Abhinav-grd:


src/pages/gallery/index.tsx
```diff
@@ -384,40 +384,71 @@ export default function Gallery(props) {
                                 }
                             });
 
+                            const loadMoreItems = async (startIndex, stopIndex) => {
```
pushkar8723, line 387: [question]
Since we have everything in local storage that too in one object, how is this helping?

Abhinav-grd: loadMoreItems loads the preview image of items by making the fetch call  not the actual items itself

pushkar8723:
use `overscanCount` prop instead.

PR #15, opened 2021-02-15
show alert message and loader on first load

Abhinav-grd:


vishnukvmd (approved):

vishnukvmd (changes requested):
The `Alert` is displayed not just on the first load, but on subsequent loads too. Maybe we could persist this state?

vishnukvmd (approved):

PR #17, opened 2021-02-15
corrected thumbnail compression

Abhinav-grd:


src/services/uploadService.ts
```diff
@@ -465,7 +465,7 @@ class UploadService {
                             canvas.height
                         );
                         var image = canvas.toDataURL();
-                        var success = image.length > 100000;
+                        var success = image.length;
```
vishnukvmd, line 468: What's the expectation out of the return value of `snapImage`? That it should be a non-zero value? I'd still recommend returning a boolean explicitly here.

vishnukvmd (approved):

vishnukvmd (approved):

PR #25, opened 2021-02-18
corrected metadataMap seeding function

Abhinav-grd:


src/services/uploadService.ts
```diff
@@ -356,38 +356,30 @@ class UploadService {
...
+                metadataJSON['geoDataExif']['latitude'] != 0.0 ||
+                metadataJSON['geoDataExif']['longitude'] != 0.0
+            ) {
+                locationData = metadataJSON['geoDataExif'];
+            }
+            if (locationData != null) {
+                metaDataObject['latitude'] = locationData['latitide'];
+                metaDataObject['longitude'] = locationData['longitude'];
             }
+            this.metadataMap.set(metadataJSON['title'], metaDataObject);
```
vishnukvmd, line 382: In a separate pull request, we should use a more unique key for this map than `title` since there could be collisions across folders?

vishnukvmd (approved):

PR #26, opened 2021-02-23
Rendering heic images

Abhinav-grd:


src/services/downloadManager.ts
```diff
@@ -65,20 +68,36 @@ class DownloadManager {
...
+                    console.log('get file failed ', e);
                 }
             })();
             this.fileDownloads.set(file.id, download);
         }
         return await this.fileDownloads.get(file.id);
     };
+
+    private async convertHEIC2JPEG(fileBlob): Promise<Blob> {
+        const heic2any = require('heic2any');
```
vishnukvmd, line 94: We could define this at the top since DownloadManager already maintains state? Will prevent reinitializations.

vishnukvmd (approved):

src/services/downloadManager.ts
Abhinav-grd: There was issue where the next build failed when i tried to load this library. The problem was that the library used some browser defined things(Blob) so, the pre-render on server side failed. 
Found a way to skip that (edit pushed in the next commit)

PR #35, opened 2021-03-08
integrated sentry

Abhinav-grd:


next.config.js
```diff
@@ -1,19 +1,68 @@
...
     enabled: process.env.ANALYZE === 'true',
 });
 
+const {
+    NEXT_PUBLIC_SENTRY_DSN: SENTRY_DSN,
+    SENTRY_ORG,
+    SENTRY_PROJECT,
+    SENTRY_AUTH_TOKEN,
+    NODE_ENV,
+    VERCEL_GITHUB_COMMIT_SHA: COMMIT_SHA,
```
pushkar8723, line 14: Will `COMMIT_SHA` be present in the env variable during build time on cloud flare?

pushkar8723 (approved):

next.config.js
Abhinav-grd: @vishnukvmd  Is there some way to get this.?

vishnukvmd: I don't see any. https://developers.cloudflare.com/pages/platform/build-configuration

Will require more digging.

For now we'll continue to use Netlify for bada-frame. They seem to have `COMMIT_REF`?

https://docs.netlify.com/configure-builds/environment-variables/

Abhinav-grd: Okk...then lets configure for netlify for now and leave the problem of configuring for cloudflare for when we migrate

PR #45, opened 2021-03-20
Implemented tile selection feature

pushkar8723:
Screenshot 2021-03-20 at 11 48 30 PM


Abhinav-grd:
The Click on the tile selection radio button should ideally not open the image but it does , are you working on that kya?

pushkar8723:
> The Click on the tile selection radio button should ideally not open the image but it does , are you working on that kya? @pushkar8723

yes I will take care of it.

PR #49, opened 2021-04-02
Intermediate key

Abhinav-grd:


src/pages/credentials/index.tsx
```diff
@@ -66,6 +68,15 @@ export default function Credentials() {
                     keyAttributes.keyDecryptionNonce,
                     kek
                 );
+                if (isFirstLogin()) {
+                    const intermediateKeyAttribute = await generateIntermediateKey(
```
vishnukvmd, line 72: `intermediateKeyAttributes`

src/utils/crypto/index.ts
```diff
@@ -0,0 +1,31 @@
+import { KEK } from 'pages/generate';
+import { B64EncryptionResult } from 'services/uploadService';
+import { KeyAttributes } from 'types';
+import CryptoWorker from './cryptoWorker';
+
+export async function generateIntermediateKey(
```
vishnukvmd, line 6: `generateIntermediateKeyAttributes` since it's returning all the attributes and not just the `intermediateKey`??

src/utils/common/utilFunctions.ts
```diff
@@ -15,3 +16,9 @@ export function getFileExtension(fileName): string {
 export function runningInBrowser() {
     return typeof window !== 'undefined';
 }
+
+export const isFirstLogin = () => getData(LS_KEYS.IS_FIRST_LOGIN)?.status;
```
vishnukvmd, line 20: Should this default to `true`?

vishnukvmd (approved):

src/utils/common/utilFunctions.ts
Abhinav-grd: ``` true```?? or ```false``` because only if a user passes through the verify page(where i set it to true)....will he be logging in for the first time

PR #104, opened 2021-07-06
Logout user from server

ua741:


src/services/userService.ts
```diff
@@ -168,3 +170,15 @@ export const getTwoFactorStatus = async () => {
...
+
+export const _logout = async () => {
+    if (!getToken()) return true;
+    try {
+        await HTTPService.post(`${ENDPOINT}/users/logout`, null, null, {
+            'X-Auth-Token': getToken(),
+        });
+        return true;
+    } catch (e) {
+        return false;
```
Abhinav-grd, line 182: we should ideally log this error using ```logError(e,message)```

Abhinav-grd (approved):

src/services/userService.ts
ua741: > we should ideally log this error using `logError(e,message)`

Good catch. Fixed it.

PR #114, opened 2021-08-09
Upload Service refactoring

Abhinav-grd:


src/utils/upload/index.ts
```diff
@@ -0,0 +1,51 @@
+import { FileWithCollection } from 'services/upload/uploadManager';
+import { FileInMemory, MetadataObject } from 'services/upload/uploadService';
+
+const TYPE_JSON = 'json';
+
+export function fileAlreadyInCollection(
+    existingFilesInCollection,
```
ua741, line 7: Is it possible to specify type of `existingFilesInCollection` similar to `newFile:` **FileInMemory**?

src/utils/network/index.ts
```diff
@@ -0,0 +1,17 @@
+import { sleep } from 'utils/common';
+
+const retrySleepTime = [2000, 5000, 10000];
+
+export async function retryAsyncFunction(func: ()=>Promise<any>, retryCount: number = 3) {
```
ua741, line 5: What would be the behavior when someone specify `retryCount  >=3`?
Also, if am not wrong, if we call the `retryAsyncFunction` with `retryCount=1`, we would end up retrying after 10000 unit of time..

src/utils/upload/index.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function segregateFiles(
+    filesWithCollectionToUpload: FileWithCollection[],
+) {
+    const metadataFiles: FileWithCollection[] = [];
+    const mediaFiles: FileWithCollection[] = [];
+    filesWithCollectionToUpload.forEach((fileWithCollection) => {
+        const file = fileWithCollection.file;
+        if (file?.name.substr(0, 1) === '.') {
```
ua741, line 40: nit: can use `startsWith('.')` for better readability.

src/services/upload/uploader.ts
```diff
@@ -0,0 +1,83 @@
...
+        if (e.message === CustomError.ETAG_MISSING) {
+            UIService.setFileProgress(rawFile.name, FileUploadResults.BLOCKED);
+            return { fileUploadResult: FileUploadResults.BLOCKED };
+        } else {
+            UIService.setFileProgress(rawFile.name, FileUploadResults.FAILED);
+            return { fileUploadResult: FileUploadResults.FAILED };
+        }
+    } finally {
+        file = null;
+        encryptedFile = null;
```
ua741, line 81: Both of these variables are defined within the scope of this function. Do we always need to set them as null for garbage collection? I am not very familiar with memory management in JS, so curious about this.

src/services/upload/uploader.ts
```diff
@@ -0,0 +1,83 @@
...
+    UIService.setFileProgress(rawFile.name, 0);
+    let file: FileInMemory = null;
+    let encryptedFile: EncryptedFile = null;
+    try {
+        file = await UploadService.readFile(reader, rawFile, collection);
+
+        if (fileAlreadyInCollection(existingFilesInCollection, file)) {
+            // set progress to -2 indicating that file upload was skipped
+            UIService.setFileProgress(rawFile.name, FileUploadResults.SKIPPED);
+            await sleep(TwoSecondInMillSeconds);
```
ua741, line 40: What's the reason for adding the 2 second sleep here? Can we add a comment mentioning the same?

src/pages/gallery/index.tsx
```diff
@@ -225,6 +225,7 @@ export default function Gallery() {
                         title: constants.SESSION_EXPIRED,
                         content: constants.SESSION_EXPIRED_MESSAGE,
                         staticBackdrop: true,
+                        nonClosable: true,
```
ua741, line 228: Assuming you meant to make this change. Doesn't look related to Upload refactoring and linting..

src/services/upload/thumbnailService.ts
```diff
@@ -0,0 +1,170 @@
+import { FILE_TYPE } from 'pages/gallery';
```
ua741, line 1: Shall we also migrate `FILE_TYPE` enum out of `pages` namespace?

src/pages/gallery/index.tsx
Abhinav-grd: yes..i should move this to master, found it while making refactoring...and commit on this branch by mistake

src/utils/upload/index.ts
```diff
@@ -0,0 +1,51 @@
...
+            // ignore files with name starting with . (hidden files)
+            return;
+        }
+        if (file.name.slice(-4) === TYPE_JSON) {
```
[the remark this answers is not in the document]
Abhinav-grd: ? is not required.  will remove it

src/utils/network/index.ts
Abhinav-grd: so we never pass the retryCount and it always takes value as three, it is set so that next recursive call could be made with retryCount-1 value.

Abhinav-grd: will add a wrapper function around it to remove the confusion

src/services/upload/uploader.ts
Abhinav-grd: there was a memory leak issue with uploader once and this solved it :sweat_smile: i think because of the recursive call made

src/services/upload/uploader.ts
Abhinav-grd: so that file remain in the progressbar screen for 2 second with the message - "duplicate file skipping..."

src/services/upload/readFileService.ts
```diff
@@ -0,0 +1,104 @@
...
+    ) {
+        fileType = FILE_TYPE.IMAGE;
+    }
+    return fileType;
+}
+
+export function getFileOriginalName(file: globalThis.File) {
+    let originalName: string = null;
+
+    const isEditedFile = file.name.endsWith(EDITED_FILE_SUFFIX);
```
ua741, line 42: Is this some common standard where the edited file.name will be something like _imageName.jpeg-edited_ ? 
Can we add some comments/documentation about this part :)

src/services/upload/readFileService.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (isEditedFile) {
+        originalName = file.name.slice(0, -1 * EDITED_FILE_SUFFIX.length);
+    } else {
+        originalName = file.name;
+    }
+    return originalName;
+}
+
+function getFileStream(reader: FileReader, file: globalThis.File) {
```
ua741, line 51: nit: We can it more generic and pass the `chunkSize` as a input to `getFileStream` and `fileChunkReaderMaker` method. This two func doesn't need to be aware of `ENCRYPTION_CHUNK_SIZE`.

src/services/upload/uploadService.ts
```diff
@@ -0,0 +1,263 @@
...
+                extractedMetadata[key] = value;
+            }
+
+            const filedata = await getFileData(reader, rawFile);
+
+            return {
+                filedata,
+                thumbnail,
+                metadata: extractedMetadata,
+            } as FileInMemory;
```
ua741, line 135: nit: should we also declare the `returnType` in `readFile` method signature?

src/services/upload/uploader.ts
```diff
@@ -0,0 +1,84 @@
...
+        );
+
+        const uploadFile: UploadFile = UploadService.getUploadFile(
+            collection,
+            backupedFile,
+            encryptedFile.fileKey
+        );
+
+        const uploadedFile = await NetworkClient.uploadFile(uploadFile);
+        const decryptedFile = await decryptFile(uploadedFile, collection);
```
ua741, line 62: What's the purpose of decrypting the file right after encrypting and uploading. Is it possible to reuse the `file` which is passed in the `UploadService.encryptFile`.  I am curious about the impact of this method call during upload of large video lines.

src/services/upload/thumbnailService.ts
```diff
@@ -0,0 +1,170 @@
...
+        } catch (e) {
+            logError(e);
+            // ignore and set staticThumbnail
+            hasStaticThumbnail = true;
+        }
+        const thumbnailBlob=await thumbnailCanvasToBlob(canvas);
+        const thumbnail = await getUint8ArrayView(
+            reader,
+            thumbnailBlob,
+        );
```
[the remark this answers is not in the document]
Abhinav-grd: okk yaa..i should intialize the canvas with an empty canvas element, 
as generating a fallback thumbnail from blank canvas which is just a complete black screen

src/services/upload/thumbnailService.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export async function thumbnailCanvasToBlob(canvas:HTMLCanvasElement) {
+    let thumbnailBlob = null;
+    let attempts = 0;
+    let quality = 1;
+
+    do {
+        attempts++;
+        quality /= 2;
```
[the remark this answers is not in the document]
Abhinav-grd: yes because the generated after the first attempt its size may not less then the desired thumbnailSize and then we attempt to reduce its quality more

src/services/upload/s3Service.ts
```diff
@@ -0,0 +1,103 @@
+import { CHUNKS_COMBINED_FOR_A_UPLOAD_PART, DataStream } from './uploadService';
```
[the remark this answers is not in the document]
Abhinav-grd: yes. i should rename it

src/services/upload/s3Service.ts
```diff
@@ -0,0 +1,103 @@
...
+            uploadChunk,
+            progressTracker,
+        );
+        partEtags.push({ PartNumber: index + 1, ETag: eTag });
+    }
+    await completeMultipartUpload(partEtags, multipartUploadURLs.completeURL);
+    return multipartUploadURLs.objectKey;
+}
+
+export function getRandomProgressPerPartUpload(uploadPartCount: number) {
```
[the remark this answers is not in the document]
Abhinav-grd: No..had forgotten to remove the export after refactoring...will remove

src/services/upload/readFileService.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export async function getFileData(reader: FileReader, file: globalThis.File) {
+    return file.size > MIN_STREAM_FILE_SIZE
+        ? getFileStream(reader, file)
+        : await getUint8ArrayView(reader, file);
+}
+
+export function getFileType(receivedFile: globalThis.File) {
+    let fileType: FILE_TYPE;
+    switch (receivedFile.type.split('/')[0]) {
```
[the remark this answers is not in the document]
Abhinav-grd: ohh will add a helper function

src/services/upload/readFileService.ts
Abhinav-grd: okk..Will add comment :smile:

src/services/upload/uploader.ts
Abhinav-grd: So we store the decrypted file Key along with file in localDB  ,and we never receive the actual file Key back in the uploader function so, we have to decrypt it using the collection key. we could have avoided decrypting the metadata but already had a util function for decryption so used it. Decrpyting metadata is cheap operation

So we are not decrypting the actual file just the metadata of the file so it should not be a problem.

ua741: Got it. I got confused. I was not aware that `uploadedFile` actually refers to `enteFile` object, not the media file.

src/services/upload/uploadService.ts
```diff
@@ -0,0 +1,259 @@
...
+    getFileData,
+} from './readFileService';
+import { encryptFiledata } from './encryptionService';
+import { ENCRYPTION_CHUNK_SIZE } from 'types';
+import { uploadStreamUsingMultipart } from './multiPartUploadService';
+import UIService from './uiService';
+import { parseError } from 'utils/common/errorUtil';
+import { MetadataMap } from './uploadManager';
+
+export const MIN_STREAM_FILE_SIZE = 20 * 1024 * 1024;
```
ua741, line 23: MULTIPART_PART_SIZE  
This is the chunk size of a file which will be uploaded in a single part.

ua741: FILE_READER_CHUNK_SIZE = ENCRYPTION_CHUNK_SIZE

src/services/upload/multiPartUploadService.ts
```diff
@@ -0,0 +1,103 @@
...
+    return percentPerPart;
+}
+
+async function combineChunksToFormUploadPart(
+    streamReader: ReadableStreamDefaultReader<Uint8Array>
+) {
+    const combinedChunks = [];
+    for (let i = 0; i < CHUNKS_COMBINED_FOR_A_UPLOAD_PART; i++) {
+        const { done, value: chunk } = await streamReader.read();
+        if (done) {
```
ua741, line 83: Just to be safe, similar to streamReader.read(), shall we also return `{Uint8Array.from(combinedChunks), done}`

We should only call  `await completeMultipartUpload(partEtags, multipartUploadURLs.completeURL);` when we are 100% sure that we have read everything from `streamReader`

ua741 (approved):

src/services/upload/uploader.ts
Abhinav-grd: will make this change on different branch will change confusion here

src/services/upload/exifService.ts
```diff
@@ -0,0 +1,100 @@
+import EXIF from 'exif-js';
+
+import { logError } from 'utils/sentry';
+import { NULL_LOCATION, Location } from './metadataService';
+
+const SOUTH_DIRECTION = 'S';
+const WEST_DIRECTION = 'W';
+const EXIF_HAVING_TYPES = new Set(['jpeg', 'jpg', 'tiff']);
```
ua741, line 8: How do we handle exif data for HEIC file types? 
Also, is this list based on our observation or it's standard list? (Want to check that we are not missing out any other common types here)

PR #124, opened 2021-08-28
Add delete account button on sidebar

ua741:
image

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -76,13 +76,14 @@ export default function Sidebar(props: Props) {
         const win = window.open(feedbackURL, '_blank');
         win.focus();
     }
-    function openSupportMail() {
-        const a = document.createElement('a');
-        a.href = 'mailto:contact@ente.io';
 
+    function initiateEmail(email: string) {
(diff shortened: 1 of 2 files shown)


src/components/Sidebar.tsx
```diff
@@ -309,6 +309,27 @@ export default function Sidebar(props: Props) {
...
+                    variant="danger"
+                    style={{ marginTop: '30px' }}
+                    onClick={() =>
+                        props.setDialogMessage({
+                            title: `${constants.DELETE_ACCOUNT}`,
+                            content: constants.DELETE_MESSAGE(),
+                            staticBackdrop: true,
+                            proceed: {
+                                text: constants.DELETE_ACCOUNT,
+                                action: () => {
```
Abhinav-grd, line 322: you can create a function for a action and call it here. 

Something similar to https://github.com/ente-io/bada-frame/blob/1d2b6e3243b4e40be548db30446fe539e3b627c5/src/components/Sidebar.tsx#L79

ua741: Done, thanks for this suggestion.

PR #139, opened 2021-09-07
Exit inactive ffmpeg

Abhinav-grd:


ua741:
Let's discuss about this offline and decide if we need this now or not.

PR #143, opened 2021-09-10
Update exif package

Abhinav-grd:


src/utils/exif/index.ts
```diff
@@ -0,0 +1,25 @@
+export function prettyPrintExif(data: Object) {
+    let a;
```
ua741, line 2: What's `a` here? Wondering if it's possible to rename it to something more readable?

src/utils/exif/index.ts
```diff
@@ -0,0 +1,25 @@
...
+                    data[a] +
+                    ' [' +
+                    data[a].numerator +
+                    '/' +
+                    data[a].denominator +
+                    ']\r\n';
+            } else {
+                strPretty += a + ' : [' + data[a].length + ' values]\r\n';
+            }
+        } else {
```
ua741, line 20: When the type isn't of `object` type, then it will be always of type `string`?

src/services/upload/exifService.ts
```diff
@@ -34,17 +35,31 @@ export async function getExifData(
...
+                Number(parts[0]),
+                Number(parts[1]) - 1,
+                Number(parts[2])
+            );
+            return date.getTime() * 1000;
+        } else {
+            throw Error(CustomError.MISSING_EXIF_TAG);
+        }
+    } catch (e) {
+        logError(e, 'getUnixTime failed', { exifData });
```
ua741, line 60: Shall we avoid explicitly logging `exifData` data as it would contain some private information about the photo?  If we feel the stack trace or exception isn't enough for debugging, we can plan to add it later.

src/services/upload/exifService.ts
```diff
@@ -34,17 +35,31 @@ export async function getExifData(
 }
 
 function getUNIXTime(exifData: any) {
-    const dateString: string = exifData.DateTimeOriginal || exifData.DateTime;
-    if (!dateString || dateString === '0000:00:00 00:00:00') {
+    try {
+        if (!exifData.DateTimeOriginal) {
+            throw Error(CustomError.MISSING_EXIF_TAG);
```
ua741, line 40: If I am understanding it right, this exception will be always logged on sentry.. Is that understanding incorrect?

src/utils/exif/index.ts
Abhinav-grd: yes..I shoud have done that...Just copied and pasted the code from a library 😬

src/services/upload/exifService.ts
Abhinav-grd: yes it will be

src/utils/exif/index.ts
Abhinav-grd: will update the code

src/utils/exif/index.ts
Abhinav-grd: any thing other than object can be printed as it is

src/services/upload/exifService.ts
ua741: Okay. This will probably result in too much noise on sentry and it's acceptable scenario where a photo doesn't have exif data.. Can we only log when something unexpected happens.. like date parsing failed etc etc.

src/services/upload/exifService.ts
```diff
@@ -1,90 +1,54 @@
...
 export async function getExifData(
-    worker,
     receivedFile: globalThis.File
 ): Promise<ParsedEXIFData> {
     try {
-        const fileChunk = await worker.getUint8ArrayView(
-            receivedFile.slice(0, CHUNK_SIZE_FOR_EXIF_READING)
-        );
-        const exifData = EXIF.readFromBinaryFile(fileChunk.buffer);
+        const exifData = await exifr.parse(receivedFile, EXIF_TAGS_NEEDED);
```
Abhinav-grd, line 24: we can pass the file directly to the package and it will read it using fileReader so custom chunking is not need

src/services/upload/exifService.ts
```diff
@@ -1,90 +1,54 @@
...
-        Number(parts[1]) - 1,
-        Number(parts[2])
-    );
-    return date.getTime() * 1000;
+    return dateTime.getTime() * 1000;
 }
 
 function getEXIFLocation(exifData): Location {
-    if (!exifData.GPSLatitude) {
+    if (!exifData.latitude || !exifData.longitude) {
```
Abhinav-grd, line 50: the package added this latitude and longitude property by parsing the GPS exif data so custom parsing is not required

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -263,16 +264,17 @@ function PhotoSwipe(props: Iprops) {
...
-                    const exif = EXIF.getAllTags(this);
-                    exif.raw = EXIF.pretty(this);
-                    if (exif.raw) {
-                        setExif(exif);
+                exifr.parse(img).then(function (exifData) {
+                    if (!exifData) {
+                        return;
+                    }
+                    exifData.raw = prettyPrintExif(exifData);
+                    if (exifData) {
```
ua741, line 276: Is there a typo here? If not, do we need this if block considering we are already checking for `if (!exifData)`?

src/services/upload/exifService.ts
ua741: Can you confirm that  `exifr.parse` won't not fail when any tag in `EXIF_TAGS_NEEDED` is missing?

Abhinav-grd: yes have tested that

src/components/PhotoSwipe/PhotoSwipe.tsx
Abhinav-grd: No, not a typo, we want to set the setExif UI state  only when exifData is not null

ua741 (approved):

ua741 (approved):

PR #145, opened 2021-09-12
Download queue

abhinavkgrd:
# Description
Add queue to thumbnail and file Download 
Have set the max concurrent download to `5` for now....can discuss and change it

Also we can decide the processing order and the order in which request are pushed into the queue....

I think handling new request first is also a option as a user may want to see the photo he is currently on rather than other that had been queued before it

# Test Plan
Tested by setting queue limit to `1` and see the network request happening

### DIFF
--- a/src/components/PhotoFrame.tsx
+++ b/src/components/PhotoFrame.tsx
@@ -271,7 +271,8 @@ const PhotoFrame = ({
             if (galleryContext.thumbs.has(item.id)) {
                 url = galleryContext.thumbs.get(item.id);
             } else {
-                url = await DownloadManager.getPreview(item);
+                const response = await DownloadManager.getThumbnail(item);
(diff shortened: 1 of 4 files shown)


ua741:
We can increase the thumbnail download to 20-30. It should be fine to download those many thumbnails in parallel

ua741:
Apart from the max concurrent download threshold, everything else looks good.  If the threshold for file will also impact the export, we should also increase that to 20(?).

Abhinav-grd:
> Apart from the max concurrent download threshold, everything else looks good. If the threshold for file will also impact the export, we should also increase that to 20(?).

No That's separate and is single threaded for now....We can update it later

ua741 (approved):

ua741 (approved):

PR #146, opened 2021-09-12
Cancel request

abhinavkgrd:
## Description
Adds logic at the caller level to cancel the request made , using the canceller provided.
Addes logic to photoswipe to call cancelRequest when user switches to a next  file for the prev file

## Test
-  check request is cancelled after the given time
- checked by switch to next video waiting for 5 sec for the first video to be canceled and then go back to see if a re-request is made
- switch between two video within 5 sec to check non on them are cancelled
- cancelled request dont trigger logError

### DIFF
--- a/src/services/downloadManager.ts
+++ b/src/services/downloadManager.ts
@@ -3,7 +3,7 @@ import { getFileUrl, getThumbnailUrl } from 'utils/common/apiUtil';
 import CryptoWorker from 'utils/crypto';
 import { generateStreamFromArrayBuffer, convertForPreview } from 'utils/file';
 import HTTPService from './HTTPService';
-import { File, FILE_TYPE } from './fileService';
+import { File } from './fileService';
 import { logError } from 'utils/sentry';
 import QueueProcessor, { RequestCanceller } from './upload/queueProcessor';
 import { CustomError } from 'utils/common/errorUtil';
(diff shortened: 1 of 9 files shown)


src/services/upload/queueProcessor.ts
```diff
@@ -50,15 +54,14 @@ export default class QueueProcessor<T> {
...
-
+            let response: any;
             if (queueItem.isCanceled.status) {
                 response = null;
             } else {
                 try {
                     response = await queueItem.request(queueItem.canceller);
                 } catch (e) {
-                    response = null;
+                    // ignore
```
ua741, line 64: What are the cases when we will get exception and why are we ignore it instead of handling/logging it?

src/services/downloadManager.ts
```diff
@@ -131,13 +136,14 @@ class DownloadManager {
         }
         if (
             file.metadata.fileType === FILE_TYPE.IMAGE ||
-            file.metadata.fileType === FILE_TYPE.LIVE_PHOTO
+            file.metadata.fileType === FILE_TYPE.LIVE_PHOTO ||
+            file.metadata.fileType === FILE_TYPE.VIDEO
```
ua741, line 140: Sorry, don't have full context but why we are adding VIDEO here now? Also, if we are adding VIDEO, why not just remove this condition and use it for all types?

src/services/upload/queueProcessor.ts
Abhinav-grd: all the exception are handled at the request level and logged there ( as this function can't bubble up the error to caller) so no error should come here because...but ya i should log the error and then ignore it

src/services/downloadManager.ts
Abhinav-grd: yes would be same....we can discuss this on call

PR #147, opened 2021-09-13
Upload behavior

Abhinav-grd:


src/components/pages/gallery/Upload.tsx
```diff
@@ -89,29 +89,29 @@ export default function Upload(props: Props) {
             },
             props.setFiles
         );
-    });
+    }, []);
```
Abhinav-grd, line 92: make the above run only one , instead of on all renders

ua741 (approved):

PR #150, opened 2021-09-16
fix looped resize

Abhinav-grd:
### Hypothesis 

`autoSizer`- react -window component for width and height detection and passing to react window)

The width of the wrapping container  of the `autoSizer`   set to auto , whereas the autoSizer sets the width for react window.

the change of width for the container is quicker than the change in value of the `autoSizer` width and height property. which tab width is changed.

So, when the width of the tab decreases then the `autoSizer` would take some time to follow and for that time the `autoSizer` was bigger than the container which caused the content to overflow and then a scrollbar to appear.

because of the scrollbar, the width of the viewport is futhur decreased which triggered another decrease event for `autoSizer`.

 But when the `autoSizer` reached that , width the content is now the size of the viewPort and so dont need for the scrollbar anymore and it disappeared.

because of which width  increases and then the autoSizer started to follow and increased its value to view port, but for some reason  when autoSizer width reaches to 100% width, a scroll bar appears 

This causes the width to again reduce and causes the ui to go into a loop

### Possible solution
 allow the container to overflow and hide the overflow, the `autoSizer` catch up in sometime and adjust to viewport

## Test Plan

Tried to resize the window seem to have worked and scrollbar is not appearing

### DIFF
--- a/src/components/PhotoFrame.tsx
+++ b/src/components/PhotoFrame.tsx
@@ -53,7 +53,7 @@ const Container = styled.div`
     width: 100%;
     flex-wrap: wrap;
     margin: 0 auto;
-    overflow: hidden;
+    overflow-x: hidden;
     .pswp-thumbnail {
         display: inline-block;
         cursor: pointer;
--- a/src/components/PhotoFrame.tsx
+++ b/src/components/PhotoFrame.tsx
@@ -53,7 +53,7 @@ const Container = styled.div`
     width: 100%;
     flex-wrap: wrap;
     margin: 0 auto;
-
+    overflow: hidden;
     .pswp-thumbnail {
         display: inline-block;
         cursor: pointer;


src/components/PhotoFrame.tsx
```diff
@@ -53,7 +53,7 @@ const Container = styled.div`
     width: 100%;
     flex-wrap: wrap;
     margin: 0 auto;
-
+    overflow: hidden;
```
pushkar8723, line 56: `overflow: scroll` would be safer bet here.

Abhinav-grd: adding`overflow :scroll` adds a scrollbar even when scrollbar is not , `overflow:auto` would be better ?

Abhinav-grd: I noticed a thing when i was trying to do this....The of the react `AutoSizer` is providing width and height value greater than that of the window because of which we get a scrollbar even with `overflow:auto`

image

Abhinav-grd: I think it would better if we put css condition to take `min (react_AutoSizer_value ,100%)`

Abhinav-grd: I think I understand what the problem is....It doesn't take into account the scrollbar width

Abhinav-grd: Ohh The value of autSizer was off by 1 px. 
This  caused a horizontal scrollbar to appear..
and because of that scrollbar the height reduced and a veritical scrollbar appears

pushkar8723 (approved):

ua741 (approved):

ua741 (approved):

pushkar8723 (approved):

pushkar8723 (approved):

PR #151, opened 2021-09-16
Move queue processor

abhinavkgrd:
## Description
Move the queueProcessor to the root of service directory. as its used in more than the upload service

## Test Plan
Builds successfully , shows that all the imports are corrected changed


abhinavkgrd:
done in the major type refactoring 

don't need it anymore

PR #154, opened 2021-09-20
Disable shared collection ops

Abhinav-grd:
## Description
Disable selecting (thereby add and move operation) and favorite button when user open shared Collection files

hides shared collection file from all section

## Test Plan

- [x] selecting disabled on shared collection
- [x] favorite icon disabled from shared files when open from shared Collection 
- [x] check shared file are not visible in all section but in shared album

### DIFF
--- a/src/utils/file/index.ts
+++ b/src/utils/file/index.ts
@@ -191,7 +191,7 @@ export async function convertForPreview(file: File, fileBlob: Blob) {
 export function isSharedFile(file: File) {
     const user: User = getData(LS_KEYS.USER);
 
-    if (!user.id || !file.ownerID) {
+    if (!user?.id || !file?.ownerID) {
         return false;
     }
     return file.ownerID !== user.id;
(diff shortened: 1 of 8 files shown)


ua741 (approved):

src/utils/file/index.ts
```diff
@@ -185,3 +187,12 @@ export async function convertForPreview(file: File, fileBlob: Blob) {
     }
     return fileBlob;
 }
+
+export function isSharedFile(file: File) {
+    const user: User = getData(LS_KEYS.USER);
+
+    if (!user.id || !file.ownerID) {
+        return false;
+    }
```
ua741, line 196: Curious about the need for such defensive check. Ideally, they should never be nil right if isSharedFile is getting called?

ua741 (approved):

src/utils/file/index.ts
Abhinav-grd: ya..ideally it should never happen... just wanted to avoid a false positive case (`undefined!==undefined` is `true`)

ua741 (approved):

PR #163, opened 2021-09-27
Handle failure

Abhinav-grd:
## Description

Changes 
- moved the syncing after add or move to collection logic to gallery helper functions
- moved logic which should occur even on failure (like selected file being unselecting and loading complete) to finally block to guarantee their execution

## Note to Reviewer

We can decide if we want to deselect selected files if the operation fails as the user may want to reattempt them and would have to re-select if we clear.

## Test Plan

Tested by blocking network calls for a specific request and seeing the behavior.

### DIFF
--- a/src/pages/gallery/index.tsx
+++ b/src/pages/gallery/index.tsx
@@ -326,8 +326,7 @@ export default function Gallery() {
                 setCollectionSelectorView,
                 selected,
                 files,
-                clearSelection,
-                syncWithRemote,
+
                 setActiveCollection,
                 collectionName,
                 collection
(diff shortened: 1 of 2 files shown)


ua741:
@Abhinav-grd  If there's any easy way to clear selection, then we should probably keep the selection state if the operation fails?

Abhinav-grd:
> @Abhinav-grd If there's any easy way to clear selection, then we should probably keep the selection state if the operation fails?

Ya its easy...I clear the selection myself. Its not cleared by its own....will just remove that clear operation

ua741 (approved):

PR #173, opened 2021-09-30
add delete option for archived files

Abhinav-grd:
## Description

added delete option when files in archive are selected

image

updated 

image

## Test Plan

- [x] delete works


ua741 (approved):
Can also add option to add to collection if it's simple change

ua741 (approved):

PR #181, opened 2021-10-07
search in all section instead of the collection user opened search on

Abhinav-grd:
## Description

Currently the files in the active collection in which search was made were only filtered and shown in search results

This update now makes searching in any search result in a global search.

### Implementation cavets
- The user is now redirected to All section after the search, should i add logic to redirect the user back to the collection he was on before the search is made.

## Test Plan

- [x] all collection files are searched and filtered through

### DIFF
--- a/src/pages/gallery/index.tsx
+++ b/src/pages/gallery/index.tsx
@@ -492,6 +492,7 @@ export default function Gallery() {
     };
 
     const updateSearch = (search: Search) => {
+        setActiveCollection(ALL_SECTION);
         setSearch(search);
         setSearchStats(null);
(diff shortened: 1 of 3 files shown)


src/pages/gallery/index.tsx
```diff
@@ -492,6 +492,7 @@ export default function Gallery() {
     };
 
     const updateSearch = (search: Search) => {
+        setActiveCollection(ALL_SECTION);
```
vishnukvmd, line 495: 😢

vishnukvmd (approved):

PR #197, opened 2021-10-27
Update creation time

Abhinav-grd:
## Description
^title

## screenshots

image

image

## Test Plan
- [x] creation time is actually updated
- [x] edited creation displayed on UI
- [x] have time limit - upper limit Today , lower limit - 01-01-1800 
- [x] merged meta only used on UI and not persisted
- [x] local version are updated so multiple update request possible before sync


ua741:
Can we also configure min:01-01-1800 and max date: today's date in the time selector?

Abhinav-grd:
> Can we also configure min:01-01-1800 and max date: today's date in the time selector?

okk ya...Will do

ua741 (approved):

PR #205, opened 2021-11-02
Export v2

abhinavkgrd:
### updates 

- The file names /collection names are appended with number suffices for duplicates
- older collection and files already exported are first migrated on a new export 
- edited time added to exif data as dateTimeOriginal
- only user files are exported
- on collection renames -> collection Folder name are renamed.

## Test Plan
- [x] only users files are exported
- [x] files are exported to correct collection if same name other collection is created
- [x] tested failed, new files and new export flows
- [x] test exif update fail cases , file with the incorrect file extension
- [x] tested collection name update works

### DIFF
--- a/src/services/exportService.ts
+++ b/src/services/exportService.ts
@@ -15,6 +15,7 @@ import {
     getOldFileMetadataSavePath,
     getExportedFiles,
     getMetadataFolderPath,
+    getCollectionsCreatedAfterLastExport,
 } from 'utils/export';
 import { retryAsyncFunction } from 'utils/network';
 import { logError } from 'utils/sentry';
@@ -40,9 +41,14 @@ import { updateFileCreationDateInEXIF } from './upload/exifService';
 import { MetadataObject } from './upload/uploadService';
 import QueueProcessor from './upload/queueProcessor';
 
+type CollectionIDPathMap = Map<number, string>;
 export interface ExportProgress {
     current: number;
     total: number;
+}
+export interface ExportedCollection {
+    collectionID: number;
+    folderPath: string;
 }
 export interface ExportStats {
     failed: number;
@@ -59,6 +65,7 @@ export interface ExportRecord {
     queuedFiles?: string[];
     exportedFiles?: string[];
     failedFiles?: string[];
+    exportedCollections?: ExportedCollection[];
 }
 export enum ExportStage {
     INIT,
@@ -127,6 +134,8 @@ class ExportService {
                 // no-export folder set
                 return;
             }
+            const user: User = getData(LS_KEYS.USER);
+
             let filesToExport: File[];
             const localFiles = await getLocalFiles();
             const userPersonalFiles = localFiles
@@ -138,15 +147,14 @@ class ExportService {
                 collections,
                 userPersonalFiles
             );
-            const user: User = getData(LS_KEYS.USER);
             const userCollections = nonEmptyCollections
                 .filter((collection) => collection.owner.id === user?.id)
                 .sort(
                     (collectionA, collectionB) =>
                         collectionA.id - collectionB.id
                 );
-            const exportRecord = await this.getExportRecord(exportDir);
             await this.migrateExport(exportDir, collections, userPersonalFiles);
+            const exportRecord = await this.getExportRecord(exportDir);
 
             if (exportType === ExportType.NEW) {
                 filesToExport = getFilesUploadedAfterLastExport(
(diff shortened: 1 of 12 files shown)


src/utils/file/index.ts
```diff
@@ -250,6 +259,10 @@ export function splitFilenameAndExtension(filename): [string, string] {
         ];
 }
 
+export function getFileExtension(filename) {
+    return splitFilenameAndExtension(filename)[1];
```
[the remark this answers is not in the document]
abhinavkgrd: it returns null

src/utils/file/index.ts
```diff
@@ -40,7 +43,13 @@ export function downloadAsFile(filename: string, content: string) {
 export async function downloadFile(file) {
     const a = document.createElement('a');
     a.style.display = 'none';
-    a.href = await DownloadManger.getFile(file);
+    const fileURL = await DownloadManger.getFile(file);
+    const fileBlob = await (await fetch(fileURL)).blob();
+    const updatedFileBlob = await updateFileCreationDateInEXIF(
+        fileBlob,
+        new Date(file.pubMagicMetadata.data.editedTime / 1000)
```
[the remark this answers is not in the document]
abhinavkgrd: oops!! will add null checks...

src/utils/export/index.ts
```diff
@@ -77,3 +93,70 @@ export const getGoogleLikeMetadataFile = (
...
+
+export const getFileMetadataSavePath = (
+    collectionFolderPath: string,
+    fileSaveName: string
+) => `${collectionFolderPath}/${METADATA_FOLDER_NAME}/${fileSaveName}.json`;
+
+export const getFileSavePath = (
+    collectionFolderPath: string,
+    fileSaveName: string
+) => `${collectionFolderPath}/${fileSaveName}`;
```
ua741, line 146: Should we handle the case where filename contains invalid characters?
https://stackoverflow.com/a/8485137/546896

src/utils/export/index.ts
```diff
@@ -77,3 +93,70 @@ export const getGoogleLikeMetadataFile = (
...
+        }
+    }
+    usedFilenamesInCollection.add(fileSaveName);
+    return fileSaveName;
+};
+
+export const getFileMetadataSavePath = (
+    collectionFolderPath: string,
+    fileSaveName: string
+) => `${collectionFolderPath}/${METADATA_FOLDER_NAME}/${fileSaveName}.json`;
```
ua741, line 141: why use filename for metadata instead of file_id? What will be the behaviour when two files have same file name?

src/utils/export/index.ts
```diff
@@ -77,3 +93,70 @@ export const getGoogleLikeMetadataFile = (
         2
     );
 };
+
+export const sanitizeName = (name: string) =>
+    name.replaceAll('/', '_').replaceAll(' ', '_');
```
ua741, line 98: [sanitizeName](https://stackoverflow.com/a/8485137/546896)

src/services/exportService.ts
```diff
@@ -304,57 +361,183 @@ class ExportService {
...
         return this.exportInProgress !== null;
     };
+
+    private async migrateExport(
+        exportDir: string,
+        currentVersion: number,
+        collections: Collection[],
+        files: File[]
+    ) {
+        if (currentVersion === 0) {
```
ua741, line 460: Would it make more sense to not take the current Version as input. We can fetch the currentVersion here and skip if it's already exported or something? Also, we can probably add some documentation/comment here about what are we doing for future :)

ua741 (approved):

src/utils/export/index.ts
abhinavkgrd: ya this seems like a more complete check...will update our `sanitizeName` function with this

src/utils/export/index.ts
abhinavkgrd: just named them as google photos names them....by appending `.json` to the filename the metadata is associated with.

the fileSaveName is the name after appending numbered suffixes , and so it takes care of multiple file with same name issue.

src/services/exportService.ts
abhinavkgrd: ya we can totally avoid passing currentVersion. Its just that we fetched the export_status.json , the caller function so just used that to avoid duplicate get calls.

yes will document it..i am thinking to keep the design similar to how DB migration work and so we would apply rolling updates.

abhinavkgrd: for now will simply comment about what the migration does and would later on add the rolling migration when we have more than one migrations

src/utils/export/index.ts
ua741: Okay. As our export is incremental, what would be the behaviour if a file is exported and the user decides to rename the file?

abhinavkgrd: we would re-export the the file as the file's updatedAt time will change
and when we check for files where are uploaded after last upload [1]. we would consider it as one because the entry for a exported file includes the `updationTime` [2], so we would no entry for this file

[1] https://github.com/ente-io/bada-frame/blob/export-v2/src/utils/export/index.ts#L23
[2] https://github.com/ente-io/bada-frame/blob/export-v2/src/utils/export/index.ts#L7

ua741 (approved):

vishnukvmd (approved):

vishnukvmd (dismissed):

vishnukvmd (approved):

PR #232, opened 2021-11-12
Add support to show mnemonic recovery key

ua741:
## Description

Title ^

## Test Plan
Verified that
- For same account, the web showed same 24 letter recovery code as on mobile.
- Able to recover account using 24 word mnemonic recovery key
- Able to recover account using old recovery code (bin2hex value)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@typescript-eslint/eslint-plugin": "^4.25.0",
     "@typescript-eslint/parser": "^4.25.0",
     "axios": "^0.21.3",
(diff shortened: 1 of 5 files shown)


abhinavkgrd (changes requested):
We use the same recovery key also for two-factor recovery. we will need to update that code too. 

https://github.com/ente-io/bada-frame/blob/master/src/pages/two-factor/recover/index.tsx#L44-L71

abhinavkgrd (approved):

PR #254, opened 2021-11-27
Download selected

abhinavkgrd:
## Description

add download option for  selected files

### DIFF
--- a/src/pages/gallery/index.tsx
+++ b/src/pages/gallery/index.tsx
@@ -549,9 +549,9 @@ export default function Gallery() {
     const downloadHelper = async () => {
         const selectedFiles = getSelectedFiles(selected, files);
         clearSelection();
-        loadingBar.current?.continuousStart();
+        !syncInProgress.current && loadingBar.current?.continuousStart();
         await downloadFiles(selectedFiles);
-        loadingBar.current.complete();
+        !syncInProgress.current && loadingBar.current.complete();
     };
 
     return (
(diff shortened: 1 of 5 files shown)


src/pages/gallery/index.tsx
```diff
@@ -545,6 +546,14 @@ export default function Gallery() {
         clearSelection();
     };
 
+    const downloadHelper = async () => {
```
vishnukvmd, line 549: Function names should ideally be a verb. Since they "do" something. In this case, it could be `downloadSelectedFiles`.

src/utils/file/index.ts
```diff
@@ -37,10 +37,12 @@ export function downloadAsFile(filename: string, content: string) {
     a.remove();
 }
 
-export async function downloadFile(file) {
+export async function downloadFile(file: File) {
     const a = document.createElement('a');
     a.style.display = 'none';
-    a.href = await DownloadManger.getFile(file);
+    a.href = URL.createObjectURL(
```
vishnukvmd, line 43: Why did you wrap this in an object URL now?

src/pages/gallery/index.tsx
vishnukvmd: Not resolved?

abhinavkgrd: ya....we discussed na.. ki there are many function which same nomencature will resolve it in a seperate PR

vishnukvmd (approved):

PR #255, opened 2021-11-29
Download manager improvements

abhinavkgrd:
## Description
 - store file that are converted for preview with `_converted` suffixed key instead of `forPreview` attribute as some file don't need conversion and hence same file can be use for for-preview and download and are unnecessarily re-downloaded 
 -  fix thumbnail in-memory cache not used

## Test Plan
 - [x]  tested that files only re-downloaded it the stored version is converted else...use the in memory cache
 - [x] checked duplicate cache request are not made

### DIFF
--- a/src/services/downloadManager.ts
+++ b/src/services/downloadManager.ts
@@ -1,7 +1,11 @@
 import { getToken } from 'utils/common/key';
 import { getFileUrl, getThumbnailUrl } from 'utils/common/apiUtil';
 import CryptoWorker from 'utils/crypto';
-import { generateStreamFromArrayBuffer, convertForPreview } from 'utils/file';
+import {
+    generateStreamFromArrayBuffer,
+    convertForPreview,
+    needsConversionForPreview,
+} from 'utils/file';
 import HTTPService from './HTTPService';
 import { File, FILE_TYPE } from './fileService';
 import { logError } from 'utils/sentry';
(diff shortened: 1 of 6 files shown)


src/services/downloadManager.ts
```diff
@@ -73,36 +80,36 @@ class DownloadManager {
...
             logError(e, 'Failed to get File');
             throw e;
         }
     };
 
     public async getCachedFile(file: File) {
-        return await this.fileObjectUrlPromise.get(
-            `${file.id}_forPreview=false`
-        );
+        return await this.fileObjectUrlPromise.get(file.id.toString());
```
vishnukvmd, line 112: Shouldn't this be  `${file.id}_converted` or `${file.id}`?

src/services/downloadManager.ts
```diff
@@ -16,21 +20,24 @@ class DownloadManager {
...
+                    const thumbnailCache = await caches.open('thumbs');
+                    const cacheResp: Response = await thumbnailCache.match(
+                        file.id.toString()
+                    );
+                    if (cacheResp) {
+                        return URL.createObjectURL(await cacheResp.blob());
+                    }
+                    return await this.downloadThumb(
+                        token,
+                        thumbnailCache,
```
vishnukvmd, line 34: Why does `downloadThumb` need `thumbnailCache`?

src/services/downloadManager.ts
abhinavkgrd: no..... we want to download the original files na? so even if their's a converted version we will have to download the original one

One thing we can do is to keep the original copy too along with converted one...but that seemed like a unneccesary thing as download is only a very rare thing a user would want to do and so keeping two copies a file doesn't seemed right....we can just download the original again if needed

src/services/downloadManager.ts
abhinavkgrd: ya its storing the downloaded result in cache...ya bad design will refactor

vishnukvmd (approved):

PR #304, opened 2022-01-14
Update plan fetching

abhinavkgrd:
## Description

check if the token exists and call user-plans API conditionally 

## Test Plan

tested by making calls with and without token

### DIFF
--- a/src/services/billingService.ts
+++ b/src/services/billingService.ts
@@ -16,10 +16,22 @@ enum PaymentActionType {
 
 class billingService {
     public async getPlans(): Promise<Plan[]> {
+        const token = getToken();
         try {
-            const response = await HTTPService.get(
-                `${ENDPOINT}/billing/plans/v2`
-            );
+            let response;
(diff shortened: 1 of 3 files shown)


src/services/billingService.ts
```diff
@@ -16,10 +16,22 @@ enum PaymentActionType {
 
 class billingService {
     public async getPlans(): Promise<Plan[]> {
+        const token = getToken();
         try {
-            const response = await HTTPService.get(
-                `${ENDPOINT}/billing/plans/v2`
-            );
+            let response;
+            if (!token) {
```
vishnukvmd, line 22: Will this ever be called?

abhinavkgrd: added that code for completeness...basically if in future we ever add `getPlans` without login the user , then this function would not fail

vishnukvmd (dismissed):

vishnukvmd (approved):

PR #318, opened 2022-01-22
fix file too large description

abhinavkgrd:
## Description

^title

## Test Plan

no test needed.


src/utils/strings/englishConstants.tsx
```diff
@@ -547,7 +547,7 @@ const englishConstants = {
     INPROGRESS_UPLOADS: 'uploads in progress',
     TOO_LARGE_UPLOADS: 'large files',
     TOO_LARGE_INFO:
-        'these files were not uploaded as they exceed the maximum size limit for your storage plan',
+        'these files were not uploaded as they exceed our maximum file size limit',
```
vishnukvmd, line 550: Are we reusing this string else where? If the size of the uploaded file is greater than available storage?

vishnukvmd (approved):

PR #322, opened 2022-01-25
disable report for local csp

abhinavkgrd:
## Description

^ title

## Test Plan

tested locally

### DIFF
--- a/configUtil.js
+++ b/configUtil.js
@@ -26,8 +26,6 @@ module.exports = {
         'base-uri ': "'self'",
         'frame-ancestors': " 'none'",
         'form-action': "'none'",
-        'report-uri': 'https://csp-reporter.ente.io',
-        'report-to': 'https://csp-reporter.ente.io',
     },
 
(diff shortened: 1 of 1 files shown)


configUtil.js
```diff
@@ -26,8 +26,6 @@ module.exports = {
         'base-uri ': "'self'",
         'frame-ancestors': " 'none'",
         'form-action': "'none'",
-        'report-uri': 'https://csp-reporter.ente.io',
```
vishnukvmd, line 29: This doesn't seem specific to local?

abhinavkgrd: these configurations are only for the local development server.
we were sending local reports to the same CSP-reporter webhook.
disabling that. as they are just adding noise to that group

abhinavkgrd: for production configs we have to _header file in /public folder

abhinavkgrd: @vishnukvmd

vishnukvmd (approved):

PR #358, opened 2022-02-05
fixes caching 

abhinavkgrd:
## Description

- add checks to throw an error if resp data is `undefined` and prevent storing it in cache
- update style of the footer
- add CF worker to set meta description tag

### DIFF
--- a/src/services/publicCollectionDownloadManager.ts
+++ b/src/services/publicCollectionDownloadManager.ts
@@ -40,9 +40,11 @@ class PublicCollectionDownloadManager {
                     if (cacheResp) {
                         console.log('cache hit', file.id);
                         const cacheBlob = await cacheResp.blob();
-                        const cacheURL = URL.createObjectURL(cacheBlob);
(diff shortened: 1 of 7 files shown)


src/services/publicCollectionDownloadManager.ts
```diff
@@ -37,9 +37,16 @@ class PublicCollectionDownloadManager {
                     const cacheResp: Response = await thumbnailCache?.match(
                         file.id.toString()
                     );
-                    console.log(cacheResp);
                     if (cacheResp) {
-                        return URL.createObjectURL(await cacheResp.blob());
+                        console.log('cache hit', file.id);
+                        const cacheBlob = await cacheResp.blob();
+                        console.log(
```
ua741, line 43: Do we want to log the entire object or just object size/identifier will be good enough.
Also, it can potentially be just tested using cloudflare pages/pr specific deployment?

ua741 (approved):

vishnukvmd (approved):

PR #359, opened 2022-02-05
Refactor metadata extraction v2

abhinavkgrd:
## Description

refactored type detection and metadata extraction to happen before the upload start and store it in a metadata map

## Test Plan

- [x] happy flow when all files are of the `supported type `and non-duplicate so all of them are uploaded , compared it with a upload that happened for the same files files have the correct metadata
- [x]  check file skipped because of matching metadata
- [x] check flow where metadata parsing fails for a file
- [x] check flow where fileType wasn't detected
- [x] checked flow when file size > MAX_FILE_SIZE limit

### DIFF
--- a/src/constants/upload/index.ts
+++ b/src/constants/upload/index.ts
@@ -39,4 +39,4 @@ export enum FileUploadResults {
     UPLOADED,
 }
 
-export const FIVE_GB_IN_BYTES = 5 * 1024 * 1024 * 1024;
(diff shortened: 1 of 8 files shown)


src/constants/upload/index.ts
```diff
@@ -38,3 +38,5 @@ export enum FileUploadResults {
     LARGER_THAN_AVAILABLE_STORAGE,
     UPLOADED,
 }
+
+export const FIVE_GB_IN_BYTES = 5 * 1024 * 1024 * 1024;
```
abhinavkgrd, line 42: rename to max file size instead

PR #391, opened 2022-02-24
Ffmpeg video metadata

abhinavkgrd:
## Description
- use FFmpeg to extract global video metadata
- also extracted location when available (found it in Apple photos exported video files)
 
## Test Plan
-  tested on the set of sample videos

### DIFF
--- a/src/services/upload/videoMetadataService.ts
+++ b/src/services/upload/videoMetadataService.ts
@@ -1,7 +1,7 @@
 import { NULL_LOCATION } from 'constants/upload';
 import ffmpegService from 'services/ffmpegService';
-import { getUNIXTime } from 'utils/upload';
-import { ParsedExtractedMetadata } from './metadataService';
+import { getUNIXTime } from 'utils/time';
+import { ParsedExtractedMetadata } from 'types/upload';
 
 enum VideoMetadata {
     CREATION_TIME = 'creation_time',
--- a/src/services/upload/videoMetadataService.ts
+++ b/src/services/upload/videoMetadataService.ts
@@ -10,6 +10,7 @@ enum VideoMetadata {
     APPLE_LIVE_PHOTO_IDENTIFIER = 'com.apple.quicktime.live-photo.auto',
     APPLE_CREATION_DATE = 'com.apple.quicktime.creationdate',
     APPLE_LOCATION_ISO = 'com.apple.quicktime.location.ISO6709',
+    LOCATION = 'location',
 }
 
 export async function getVideoMetadata(file: File) {
@@ -36,7 +37,8 @@ export function parseFFmpegExtractedMetadata(encodedMetadata: Uint8Array) {
     const metadataMap = Object.fromEntries(validKeyValuePairs);
 
     const location = parseAppleISOLocation(
-        metadataMap[VideoMetadata.APPLE_LOCATION_ISO]
+        metadataMap[VideoMetadata.APPLE_LOCATION_ISO] ??
+            metadataMap[VideoMetadata.LOCATION]
     );
 
     const creationTime = parseCreationTime(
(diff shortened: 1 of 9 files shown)


src/utils/time/index.ts
```diff
@@ -0,0 +1,11 @@
+export function getUNIXTime(dateTime: Date) {
```
ua741, line 1: Should we call this `getUnixTimeInMicroSeconds`?

ua741, line 1: nit: I am not familiar with the coding guidelines for the project/framework but do we really need to name such utils file like  `src/utils/time/index.ts`?  Why not just name it like `src/utils/time.js`

src/utils/time/index.ts
```diff
@@ -0,0 +1,11 @@
...
+    if (!dateTime || isNaN(dateTime.getTime())) {
+        return null;
+    }
+    const unixTime = dateTime.getTime() * 1000;
+    if (unixTime <= 0) {
+        return null;
```
ua741, line 7: iirc, unixTime will be negative for date < 1,1,1961

src/types/upload/index.ts
```diff
@@ -130,3 +130,8 @@ export interface UploadFile extends BackupedFile {
     encryptedKey: string;
     keyDecryptionNonce: string;
 }
+
+export interface ParsedExtractedMetadata {
+    location: Location;
+    creationTime: number;
```
ua741, line 136: Should we mark these fields as nullable/optional?
location`?`,
creationTime`?`

Also, instead of  `ParsedExtractMetadata`, maybe we can rename it as `ExtractedVideoMetadata`?

src/services/upload/videoMetadataService.ts
```diff
@@ -0,0 +1,74 @@
...
+    );
+    const validKeyValuePairs = metadataKeyValueArray.filter(
+        (keyValueArray) => keyValueArray.length === 2
+    ) as Array<[string, string]>;
+
+    const metadataMap = Object.fromEntries(validKeyValuePairs);
+
+    const location = parseAppleISOLocation(
+        metadataMap[VideoMetadata.APPLE_LOCATION_ISO]
+    );
```
ua741, line 40: add fallback to metadataMap['location'] if `APPLE_LOCATION_ISO` is missing?

src/utils/time/index.ts
abhinavkgrd: 😅 don't know its coding guidelines, but have seen this pattern in many javascript projects.

src/utils/time/index.ts
abhinavkgrd: This is intentional. we did it some time back when  we were getting a negative date Time value of a recent file and decided to ignore negative values all together

src/types/upload/index.ts
abhinavkgrd: 1. There were two approaches 

either I let them be optional and use `?.` whenever accessing them to check to avoid accessing a property when the value is undefined 

or make sure that they are never undefined and return a null value when let's say we don't get the data 
this way I can avoid the conditional accessing `?.` 

I have taken the second route as lit looked  cleaner

2.  `ParsedExtractMetadata` type is used for both the metadata extracted from EXIF and video metadata

src/services/upload/videoMetadataService.ts
abhinavkgrd: normal metadata don't have location data. It's an Apple-specific property

src/utils/time/index.ts
ua741: cool.  By any chance do you have the file or remember why it was happening? Might as well add a reference to that PR here so that others don't get confused.   cc @vishnukvmd

src/services/upload/videoMetadataService.ts
ua741: https://trac.ffmpeg.org/ticket/4209 This ticket does contain `location` tag. May be older videos from Apple contain location information with this key?

ua741 (approved):

ua741 (approved):

PR #392, opened 2022-02-25
Live photo playback support

rushitote:
## Description

Added encoding for live photos using FFMPEG but currently just plays the video file. So, need to find a way to play only on hover and also a loading spinner while the file is downloading/encoding.


rushitote:
Looks like this now:

image

src/components/PhotoFrame.tsx
```diff
@@ -241,6 +242,28 @@ const PhotoFrame = ({
         }
     }, [open]);
 
+    useEffect(() => {
+        const livePhotoVideo: HTMLVideoElement =
+            document.querySelector('video');
```
abhinavkgrd, line 247: there can be multiple video elements na . Does this ensure that we get the selector the current opened one ?

src/components/PhotoFrame.tsx
```diff
@@ -241,6 +242,28 @@ const PhotoFrame = ({
...
+                livePhotoVideo.play().catch(() => {
+                    livePhotoVideo.pause();
+                });
+            });
+            livePhotoBtn.addEventListener('click', function () {
+                livePhotoVideo.play().catch(() => {
+                    livePhotoVideo.pause();
+                });
+            });
+        }
```
abhinavkgrd, line 264: we should also  remove the event listeners when the component dismounts using cleanup function
https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

src/services/ffmpegService.ts
```diff
@@ -56,6 +56,38 @@ class FFmpegService {
...
+        if (!this.ffmpeg) {
+            await this.init();
+        }
+        if (this.isLoading) {
+            await this.isLoading;
+        }
+
+        try {
+            this.ffmpeg.FS('writeFile', 'input.mov', file);
+            console.log('starting encoding', new Date().toLocaleTimeString());
```
abhinavkgrd, line 70: we should avoid logging things to the console. It's ok during testing but don't forget to remove them before deployment

src/services/ffmpegService.ts
```diff
@@ -56,6 +56,38 @@ class FFmpegService {
             }
         }
     }
+
+    async convertLivePhotoToMP4(file: Uint8Array): Promise<Uint8Array> {
+        if (!this.ffmpeg) {
```
abhinavkgrd, line 61: FFmpeg can execute only one command at a time if the user switches between two live photos this function will be called for each of them and will throw an error.

use the queue processor to queue up the task and wait on the returned promise

src/services/ffmpegService.ts
```diff
@@ -56,6 +56,38 @@ class FFmpegService {
...
+        if (this.isLoading) {
+            await this.isLoading;
+        }
+
+        try {
+            this.ffmpeg.FS('writeFile', 'input.mov', file);
+            console.log('starting encoding', new Date().toLocaleTimeString());
+            await this.ffmpeg.run(
+                '-i',
+                'input.mov',
```
abhinavkgrd, line 73: try to keep the code generic so that we could reuse it to convert any type of file to any required format

src/utils/file/index.ts
```diff
@@ -333,7 +333,7 @@ export async function convertForPreview(file: EnteFile, fileBlob: Blob) {
     if (file.metadata.fileType === FILE_TYPE.LIVE_PHOTO) {
         const originalName = fileNameWithoutExtension(file.metadata.title);
         const motionPhoto = await decodeMotionPhoto(fileBlob, originalName);
-        fileBlob = new Blob([motionPhoto.image]);
+        fileBlob = new Blob([motionPhoto.video]);
```
abhinavkgrd, line 336: so the blob just has the video? when it's paused we wanted to show the motion photo image?

src/services/motionPhotoService.ts
```diff
@@ -25,8 +26,9 @@ export const decodeMotionPhoto = async (
         } else if (zipFilename.startsWith('video')) {
             motionPhoto.videoNameTitle =
                 originalName + fileExtensionWithDot(zipFilename);
-            motionPhoto.video = await zip.files[zipFilename].async(
-                'uint8array'
+            const video = await zip.files[zipFilename].async('uint8array');
+            motionPhoto.video = await FFmpegService.convertLivePhotoToMP4(
```
abhinavkgrd, line 30: we should do it in the `convertForPreview`, as this function is also used when the user clicks to download the file and in that case, we want to download the  original file without conversion

src/components/PhotoFrame.tsx
```diff
@@ -297,6 +321,38 @@ const PhotoFrame = ({
...
+                    <svg
+                        xmlns="http://www.w3.org/2000/svg"
+                        height="25px"
+                        width="25px"
+                        fill="currentColor">
+                        <path d="M0 0h24v24H0V0z" fill="none" />
+                        <path d="M6.76 4.84l-1.8-1.79-1.41 1.41 1.79 1.79zM1 10.5h3v2H1zM11 .55h2V3.5h-2zm8.04 2.495l1.408 1.407-1.79 1.79-1.407-1.408zm-1.8 15.115l1.79 1.8 1.41-1.41-1.8-1.79zM20 10.5h3v2h-3zm-8-5c-3.31 0-6 2.69-6 6s2.69 6 6 6 6-2.69 6-6-2.69-6-6-6zm0 10c-2.21 0-4-1.79-4-4s1.79-4 4-4 4 1.79 4 4-1.79 4-4 4zm-1 4h2v2.95h-2zm-7.45-.96l1.41 1.41 1.79-1.8-1.41-1.41z" />
+                    </svg> LIVE 
+                    </button>
+                    <video class = "live-photo" loop>
```
abhinavkgrd, line 338: i think we should also mute the video

src/components/PhotoFrame.tsx
```diff
@@ -250,7 +273,8 @@ const PhotoFrame = ({
             h: window.innerHeight,
         };
         if (
-            files[index].metadata.fileType === FILE_TYPE.VIDEO &&
+            (files[index].metadata.fileType === FILE_TYPE.VIDEO ||
+                files[index].metadata.fileType === FILE_TYPE.LIVE_PHOTO) &&
```
abhinavkgrd, line 277: I think we should add the live photo indicator during loading too?

abhinavkgrd:
also, merge the master to this branch. so the network request failing issue is solved.

src/components/PhotoFrame.tsx
rushitote: Sure, will add it. I had thought that removal of the HTML element itself should also remove the event listeners associated with it.

src/components/PhotoFrame.tsx
rushitote: Sure, will do that.

src/utils/file/index.ts
rushitote: The paused state is the start frame of the video which can differ from the image. Will change this.

src/services/motionPhotoService.ts
rushitote: Oh right, didn't notice this.

src/components/PhotoFrame.tsx
```diff
@@ -454,9 +533,32 @@ const PhotoFrame = ({
...
+                }
+                if (item.metadata.fileType === FILE_TYPE.LIVE_PHOTO) {
+                    let imageURL;
+                    let videoURL;
+                    if (galleryContext.files.has(item.id)) {
+                        [imageURL, videoURL] = galleryContext.files
+                            .get(item.id)
+                            .split(',');
+                    } else {
+                        ({ imageURL, videoURL } = await splitLivePhoto(
```
abhinavkgrd, line 548: we can have this logic directly in the `convertForPreview` function? 

why send the zip blob and then do the call the conversion logic here?

src/services/ffmpegService.ts
```diff
@@ -9,7 +9,7 @@ class FFmpegService {
     private isLoading = null;
     private fileReader: FileReader = null;
 
-    private generateThumbnailProcessor = new QueueProcessor<Uint8Array>(1);
+    private generateQueueProcessor = new QueueProcessor<Uint8Array>(1);
```
abhinavkgrd, line 12: let's call this `ffmpegTaskQueue`

src/utils/file/index.ts
```diff
@@ -333,9 +334,7 @@ export function generateStreamFromArrayBuffer(data: Uint8Array) {
 
 export async function convertForPreview(file: EnteFile, fileBlob: Blob) {
     if (file.metadata.fileType === FILE_TYPE.LIVE_PHOTO) {
-        const originalName = fileNameWithoutExtension(file.metadata.title);
-        const motionPhoto = await decodeMotionPhoto(fileBlob, originalName);
-        fileBlob = new Blob([motionPhoto.image]);
+        return fileBlob;
```
abhinavkgrd, line 337: It would be better if you transcode the live photo assets  here itself and then return an array of file blobs [ imageBlob, videoBlob] 

returning the `fileBlob` without conversion can later cause confusions

src/utils/photoFrame/index.ts
```diff
@@ -13,3 +13,14 @@ export async function isPlaybackPossible(url: string): Promise<boolean> {
         video.src = url;
     });
 }
+
+export const livePhotoBtnHTML = `
```
abhinavkgrd, line 17: we should keep `html` code in the component directory

you can create a file `component/livePhotoBtn` and put it there

src/services/ffmpegService.ts
```diff
@@ -101,4 +129,28 @@ async function generateThumbnailHelper(
...
+    try {
+        ffmpeg.FS('writeFile', inputFileName, file);
+        await ffmpeg.run(
+            '-i',
+            inputFileName,
+            '-preset',
+            'ultrafast',
+            'output.mp4'
+        );
+        const convertedFile = await ffmpeg.FS('readFile', 'output.mp4');
```
abhinavkgrd, line 146: don't need to await on this

src/components/PhotoFrame.tsx
```diff
@@ -324,21 +353,15 @@ const PhotoFrame = ({
...
-                        width="25px"
-                        fill="currentColor">
-                        <path d="M0 0h24v24H0V0z" fill="none" />
-                        <path d="M6.76 4.84l-1.8-1.79-1.41 1.41 1.79 1.79zM1 10.5h3v2H1zM11 .55h2V3.5h-2zm8.04 2.495l1.408 1.407-1.79 1.79-1.407-1.408zm-1.8 15.115l1.79 1.8 1.41-1.41-1.8-1.79zM20 10.5h3v2h-3zm-8-5c-3.31 0-6 2.69-6 6s2.69 6 6 6 6-2.69 6-6-2.69-6-6-6zm0 10c-2.21 0-4-1.79-4-4s1.79-4 4-4 4 1.79 4 4-1.79 4-4 4zm-1 4h2v2.95h-2zm-7.45-.96l1.41 1.41 1.79-1.8-1.41-1.41z" />
-                    </svg> LIVE 
+                        ${livePhotoBtnHTML}
                     </button>
-                    <video class = "live-photo" loop>
-                        <source src="${url}" />
+                    <video class = "live-photo" poster = "${imageURL}" loop muted>
```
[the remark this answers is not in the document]
rushitote: Yeah, was thinking on this.

rushitote:
Looks like this now:


https://user-images.githubusercontent.com/31135699/156173965-541935cf-369f-4dce-8ade-5ea762d76ebc.mp4

src/components/PhotoFrame.tsx
```diff
@@ -517,48 +564,37 @@ const PhotoFrame = ({
         if (!fetching[item.dataIndex]) {
             try {
                 fetching[item.dataIndex] = true;
-                let url: string;
+                let URLs: string[];
```
abhinavkgrd, line 542: so acronyms are capitalised only when they are not the first word. 😅  

so it should be written `urls` 😅  and capitalised,

because first letter capital is used to signify that its some sort of constant.

rushitote: Oh, didn't knew this.

src/components/PhotoFrame.tsx
```diff
@@ -392,17 +339,17 @@ const PhotoFrame = ({
                 files[index].html = `
                 <div class="video-loading">
                     <img src="${files[index].msrc}" />
-                    <div class="download-message">
+                    <div class="download-message" data-id="${files[index].id}">
                         ${constants.VIDEO_PLAYBACK_FAILED_DOWNLOAD_INSTEAD}
                         <button class = "btn btn-outline-success">Download</button>
```
abhinavkgrd, line 340: wouldn't attaching an onClick callback here itself be an easier solution. Did it not work for some reason?

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -527,6 +543,100 @@ function PhotoSwipe(props: Iprops) {
...
+                        );
+                    };
+                }
+            }
+        }
+
+        return () => {
+            setLivePhotoBtn(null);
+        };
+    }, [photoSwipe?.currItem, isOpen, isSourceLoaded]);
```
abhinavkgrd, line 638: do we need isSourceLoaded here as the `photoSwipe?.curr.item` is anyways updated when the URL is changed. ?

Is there some case you came across when it doesn't happen ?

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -527,6 +543,100 @@ function PhotoSwipe(props: Iprops) {
...
+
+            const pauseVideo = async (livePhotoVideo, livePhotoImage) => {
+                if (!videoPlaying) return;
+                videoPlaying = false;
+                livePhotoVideo.pause();
+                livePhotoVideo.style.opacity = 0;
+                livePhotoImage.style.opacity = 1;
+            };
+            setLivePhotoBtn(
+                <LivePhotoBtn
```
abhinavkgrd, line 571: I think instead of updating the complete button we can just update the eventHandler functions?

something like `fav button`. have a state isLivePhoto state which would update whenever the currentItem change and that could handle the hiding and showing of the button and just pass the current item id to play and pause Video handlers so that they could construct the className and select the element and trigger the required action

src/components/PhotoFrame.tsx
rushitote: It showed an error that you can't directly add Javascript in HTML (I don't remember the exact error, will let you know).

src/components/PhotoSwipe/PhotoSwipe.tsx
rushitote: Yeah, it doesn't seem to trigger the useEffect when the `item.html` or `item.url` changes, but it was good when the current item is changed though.

src/components/PhotoSwipe/PhotoSwipe.tsx
rushitote: Sure, will do this.

src/components/PhotoSwipe/PhotoSwipe.tsx
abhinavkgrd: I think this is fixed in the fix loading PR I have opened...will check after that... can ignore for now

rushitote: We still need that, if `isSourceLoaded` is removed the download button is not updated in case of no video playback.

vishnukvmd:
@rushitote Could you keep the playback button at the bottom right corner?

rushitote:
@vishnukvmd done in latest commit.

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -839,7 +837,17 @@ function PhotoSwipe(props: Iprops) {
...
+                    <LivePhotoBtn
+                        onClick={livePhotoBtnOptions.click}
+                        onMouseEnter={livePhotoBtnOptions.show}
+                        onMouseLeave={livePhotoBtnOptions.hide}
+                        style={{
+                            display: livePhotoBtnOptions.visible
+                                ? 'block'
+                                : 'none',
+                        }}>
+                        {livePhotoBtnHTML} LIVE
```
abhinavkgrd, line 849: we should shift this text to `englishConstants` File

context -> we keep all the constants in the `englishConstants` file. So that in future we can support localisation with minimal code changes.

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -568,76 +577,65 @@ function PhotoSwipe(props: Iprops) {
...
+                        playVideo(video, image);
+                    }
+                },
+                visible: true,
+            });
+
+            const downloadLivePhotoBtn = document.getElementById(
+                `download-btn-${item.id}`
+            ) as HTMLButtonElement;
+            console.log(downloadLivePhotoBtn);
```
abhinavkgrd, line 616: oopsie 😅

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -528,6 +553,91 @@ function PhotoSwipe(props: Iprops) {
...
+                )[0];
+                const image = document.getElementsByClassName(
+                    `live-photo-image-${item.id}`
+                )[0];
+                return { video, image };
+            };
+
+            setLivePhotoBtnOptions({
+                click: () => {
+                    const { video, image } = getvideoAndImage();
```
abhinavkgrd, line 593: I was just thinking if we need to get the image and video element each time, or we could just get it once while initialising the click event and that would do ?

src/components/PhotoSwipe/PhotoSwipe.tsx
rushitote: ah again 😶

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -528,6 +553,91 @@ function PhotoSwipe(props: Iprops) {
...
+                videoPlaying = true;
+                livePhotoVideo.style.opacity = 1;
+                livePhotoImage.style.opacity = 0;
+                livePhotoVideo.load();
+                livePhotoVideo.play().catch(() => {
+                    pauseVideo(livePhotoVideo, livePhotoImage);
+                });
+            };
+
+            const pauseVideo = async (livePhotoVideo, livePhotoImage) => {
```
abhinavkgrd, line 573: we can move pauseVideo and playVideo function outside the `useEffect` i think to util i think

src/components/PhotoSwipe/PhotoSwipe.tsx
```diff
@@ -568,76 +577,65 @@ function PhotoSwipe(props: Iprops) {
...
-                    const downloadLivePhoto = () => {
-                        downloadFileHelper(photoSwipe.currItem);
-                    };
-
-                    downloadLivePhotoBtn.addEventListener(
+            const getvideoAndImage = () => {
+                const video = document.getElementsByClassName(
+                    `live-photo-video-${item.id}`
+                )[0];
+                const image = document.getElementsByClassName(
```
abhinavkgrd, line 585: also, we can change them to ids too?

src/components/PhotoSwipe/PhotoSwipe.tsx
abhinavkgrd: did you test it out after merging master? does it work without `isSourceLoaded` now?

src/components/PhotoSwipe/PhotoSwipe.tsx
rushitote: Will try this.

src/components/PhotoSwipe/PhotoSwipe.tsx
rushitote: Yes, I did. The video and image element works without that but not the download button.

abhinavkgrd: cools... let's keep it then

abhinavkgrd (commented):
cools everything looks good to me 😄 

will let @vishnukvmd  and @ua741  test it once and then will approve it

abhinavkgrd (approved):
Go ahead @rushitote  merge it 😄

PR #444, opened 2022-03-13
Recovering failed imports on the desktop app

rushitote:
## Description
- Draft working version for recovering previously failed imports in the desktop app.

## Test Plan
- Tested uploading single/multiple files and folders and interrupting them midway.
- There is currently an issue with the timestamp of the file. Sometimes the timestamp reported by File instance differs from that of reported by the Node file stats by +/- 1ms. And so few files aren't skipped and get duplicated.

### DIFF
--- a/src/components/pages/gallery/Upload.tsx
+++ b/src/components/pages/gallery/Upload.tsx
@@ -14,6 +14,8 @@ import { logError } from 'utils/sentry';
 import { FileRejection } from 'react-dropzone';
 import UploadManager from 'services/upload/uploadManager';
 import uploadManager from 'services/upload/uploadManager';
+import ImportService from 'services/importService';
+import isElectron from 'is-electron';
 import { METADATA_FOLDER_NAME } from 'constants/export';
 import { getUserFacingErrorMessage } from 'utils/error';
 import { Collection } from 'types/collection';
@@ -87,6 +89,15 @@ export default function Upload(props: Props) {
     }, []);
 
     useEffect(() => {
+        isElectron() &&
+            ImportService.getIfToUploadFilesExists().then((exists) => {
+                if (exists) {
+                    uploadFailedFiles();
+                }
+            });
+    }, []);
+
+    useEffect(() => {
         if (
             props.acceptedFiles?.length > 0 ||
             appContext.sharedFiles?.length > 0
@@ -119,6 +130,22 @@ export default function Upload(props: Props) {
         setPercentComplete(0);
         props.closeCollectionSelector();
         setProgressView(true);
+    };
+
+    const uploadFailedFiles = async () => {
+        try {
+            console.log('Uploading failed files');
+            uploadInit();
+            const { files, collections } =
+                await ImportService.getToUploadFiles();
+            await uploadFiles(files as any, collections as any);
+        } catch (e) {
+            logError(e, 'Failed to upload previously failed files');
+        } finally {
+            // set state to done for now even if failed else
+            // user could be stuck in upload view on every restart
+            await ImportService.setDoneUploadingFiles();
+        }
     };
 
     function analyseUploadFiles(): AnalysisResult {
@@ -245,10 +272,17 @@ export default function Upload(props: Props) {
             props.setUploadInProgress(true);
             props.closeCollectionSelector();
             await props.syncWithRemote(true, true);
+            // set state to not done
+            await ImportService.setToUploadFiles(
+                filesWithCollectionToUpload,
+                collections
+            );
             await uploadManager.queueFilesForUpload(
                 filesWithCollectionToUpload,
                 collections
             );
+            // set state to done
+            await ImportService.setDoneUploadingFiles();
         } catch (err) {
             const message = getUserFacingErrorMessage(
                 err.message,
--- a/src/services/importService.ts
+++ b/src/services/importService.ts
@@ -1,0 +1,45 @@
+import { Collection } from 'types/collection';
+import { FileWithCollection } from 'types/upload';
+import { runningInBrowser } from 'utils/common';
+
+class ImportService {
+    ElectronAPIs: any;
+    private allElectronAPIsExist: boolean = false;
+
+    constructor() {
+        this.ElectronAPIs = runningInBrowser() && window['ElectronAPIs'];
+        this.allElectronAPIsExist = !!this.ElectronAPIs?.exists;
+    }
+
+    async setToUploadFiles(
+        filesWithCollectionToUpload: FileWithCollection[],
+        collections?: Collection[]
+    ) {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.setToUploadFiles(
+                filesWithCollectionToUpload,
+                collections,
+                false
+            );
+        }
+    }
+
+    async setDoneUploadingFiles() {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.setToUploadFiles([], [], true);
+        }
+    }
+
+    async getToUploadFiles() {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.getToUploadFiles();
+        }
+    }
+
+    async getIfToUploadFilesExists() {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.getIfToUploadFilesExists();
+        }
+    }
+}
+export default new ImportService();
--- a/src/services/typeDetectionService.ts
+++ b/src/services/typeDetectionService.ts
@@ -1,5 +1,5 @@
 import { FILE_TYPE } from 'constants/file';
-import { FileTypeInfo } from 'types/upload';
+import { ElectronFile, FileTypeInfo } from 'types/upload';
 import { CustomError } from 'utils/error';
 import { getFileExtension } from 'utils/file';
 import { logError } from 'utils/sentry';
@@ -18,7 +18,7 @@ export const FORMATS_TO_SKIP_TYPE_DETECTION = [
 
 export async function getFileType(
     reader: FileReader,
-    receivedFile: File
+    receivedFile: File | ElectronFile
 ): Promise<FileTypeInfo> {
     const fileExtension = getFileExtension(receivedFile.name);
     try {
@@ -29,8 +29,16 @@ export async function getFileType(
             return formatToBeSkipped;
         }
         let fileType: FILE_TYPE;
-        const typeResult = await extractFileType(reader, receivedFile);
-        const mimTypeParts = typeResult.mime?.split('/');
+        let mimTypeParts: string[];
+        let ext: string;
+        if (receivedFile instanceof File) {
+            const typeResult = await extractFileType(reader, receivedFile);
+            mimTypeParts = typeResult.mime?.split('/');
+            ext = typeResult.ext;
+        } else {
+            mimTypeParts = receivedFile.type.mimeType.split('/');
+            ext = receivedFile.type.ext;
+        }
         if (mimTypeParts?.length !== 2) {
             throw Error(CustomError.TYPE_DETECTION_FAILED);
         }
@@ -44,7 +52,7 @@ export async function getFileType(
             default:
                 fileType = FILE_TYPE.OTHERS;
         }
-        return { fileType, exactType: typeResult.ext };
+        return { fileType, exactType: ext };
     } catch (e) {
         logError(e, CustomError.TYPE_DETECTION_FAILED, {
             fileExtension,
--- a/src/services/upload/fileService.ts
+++ b/src/services/upload/fileService.ts
@@ -9,29 +9,34 @@ import {
     FileWithMetadata,
     ParsedMetadataJSONMap,
     DataStream,
+    ElectronFile,
 } from 'types/upload';
 import { splitFilenameAndExtension } from 'utils/file';
 import { logError } from 'utils/sentry';
 import { getFileNameSize, logUploadInfo } from 'utils/upload';
 import { encryptFiledata } from './encryptionService';
 import { extractMetadata, getMetadataJSONMapKey } from './metadataService';
-import { getFileStream, getUint8ArrayView } from '../readerService';
+import {
+    getFileStream,
+    getElectronFileStream,
+    getUint8ArrayView,
+} from '../readerService';
 import { generateThumbnail } from './thumbnailService';
 
 const EDITED_FILE_SUFFIX = '-edited';
 
-export function getFileSize(file: File) {
+export function getFileSize(file: File | ElectronFile) {
     return file.size;
 }
 
-export function getFilename(file: File) {
+export function getFilename(file: File | ElectronFile) {
     return file.name;
 }
 
 export async function readFile(
     reader: FileReader,
     fileTypeInfo: FileTypeInfo,
-    rawFile: File
+    rawFile: File | ElectronFile
 ): Promise<FileInMemory> {
     const { thumbnail, hasStaticThumbnail } = await generateThumbnail(
         reader,
@@ -40,7 +45,9 @@ export async function readFile(
     );
     logUploadInfo(`reading file datal${getFileNameSize(rawFile)} `);
     let filedata: Uint8Array | DataStream;
-    if (rawFile.size > MULTIPART_PART_SIZE) {
+    if (!(rawFile instanceof File)) {
+        filedata = await getElectronFileStream(rawFile, FILE_READER_CHUNK_SIZE);
+    } else if (rawFile.size > MULTIPART_PART_SIZE) {
         filedata = getFileStream(reader, rawFile, FILE_READER_CHUNK_SIZE);
     } else {
         filedata = await getUint8ArrayView(reader, rawFile);
@@ -57,7 +64,7 @@ export async function readFile(
 
 export async function extractFileMetadata(
     parsedMetadataJSONMap: ParsedMetadataJSONMap,
-    rawFile: File,
+    rawFile: File | ElectronFile,
     collectionID: number,
     fileTypeInfo: FileTypeInfo
 ) {
@@ -121,7 +128,10 @@ export async function encryptFile(
     Get the original file name for edited file to associate it to original file's metadataJSON file 
     as edited file doesn't have their own metadata file
 */
-function getFileOriginalName(file: File) {
+function getFileOriginalName(file: File | ElectronFile) {
+    if (!(file instanceof File)) {
+        return file.name;
+    }
     let originalName: string = null;
     const [nameWithoutExtension, extension] = splitFilenameAndExtension(
         file.name
--- a/src/services/upload/metadataService.ts
+++ b/src/services/upload/metadataService.ts
@@ -7,6 +7,7 @@ import {
     Location,
     FileTypeInfo,
     ParsedExtractedMetadata,
+    ElectronFile,
 } from 'types/upload';
 import { NULL_EXTRACTED_METADATA, NULL_LOCATION } from 'constants/upload';
 import { splitFilenameAndExtension } from 'utils/file';
@@ -26,11 +27,20 @@ const NULL_PARSED_METADATA_JSON: ParsedMetadataJSON = {
 };
 
 export async function extractMetadata(
-    receivedFile: File,
+    receivedFile: File | ElectronFile,
     fileTypeInfo: FileTypeInfo
 ) {
     let extractedMetadata: ParsedExtractedMetadata = NULL_EXTRACTED_METADATA;
     if (fileTypeInfo.fileType === FILE_TYPE.IMAGE) {
+        if (!(receivedFile instanceof File)) {
+            receivedFile = new File(
+                [await receivedFile.toBlob()],
+                receivedFile.name,
+                {
+                    lastModified: receivedFile.lastModified,
+                }
+            );
+        }
         extractedMetadata = await getExifData(receivedFile, fileTypeInfo);
     } else if (fileTypeInfo.fileType === FILE_TYPE.VIDEO) {
         logUploadInfo(
@@ -66,9 +76,15 @@ export const getMetadataJSONMapKey = (
 
 export async function parseMetadataJSON(
     reader: FileReader,
-    receivedFile: File
+    receivedFile: File | ElectronFile
 ) {
     try {
+        if (!(receivedFile instanceof File)) {
+            receivedFile = new File(
+                [await receivedFile.toBlob()],
+                receivedFile.name
+            );
+        }
         const metadataJSON: object = await new Promise((resolve, reject) => {
             reader.onabort = () => reject(Error('file reading was aborted'));
             reader.onerror = () => reject(Error('file reading has failed'));
@@ -79,7 +95,7 @@ export async function parseMetadataJSON(
                         : reader.result;
                 resolve(JSON.parse(result));
             };
-            reader.readAsText(receivedFile);
+            reader.readAsText(receivedFile as File);
         });
 
         const parsedMetadataJSON: ParsedMetadataJSON =
--- a/src/services/upload/videoMetadataService.ts
+++ b/src/services/upload/videoMetadataService.ts
@@ -1,9 +1,16 @@
 import { NULL_EXTRACTED_METADATA } from 'constants/upload';
 import ffmpegService from 'services/ffmpeg/ffmpegService';
+import { ElectronFile } from 'types/upload';
 import { logError } from 'utils/sentry';
 
-export async function getVideoMetadata(file: File) {
+export async function getVideoMetadata(file: File | ElectronFile) {
     let videoMetadata = NULL_EXTRACTED_METADATA;
+    if (!(file instanceof File)) {
+        file = new File([await file.toBlob()], file.name, {
+            lastModified: file.lastModified,
+            type: file.type.mimeType,
+        });
+    }
     try {
         videoMetadata = await ffmpegService.extractMetadata(file);
     } catch (e) {
--- a/src/types/upload/index.ts
+++ b/src/types/upload/index.ts
@@ -67,14 +67,29 @@ export interface ProgressUpdater {
     setHasLivePhotos: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
+export interface ElectronFile {
+    name: string;
+    path: string;
+    size: number;
+    lastModified: number;
+    type: {
+        mimeType: string;
+        ext: string;
+    };
+    createReadStream: () => Promise<ReadableStream<Uint8Array>>;
+    toBlob: () => Promise<Blob>;
+    toUInt8Array: () => Promise<Uint8Array>;
+}
+
 export interface UploadAsset {
     isLivePhoto?: boolean;
-    file?: File;
+    file?: File | ElectronFile;
     livePhotoAssets?: LivePhotoAssets;
+    isElectron?: boolean;
 }
 export interface LivePhotoAssets {
-    image: globalThis.File;
-    video: globalThis.File;
+    image: globalThis.File | ElectronFile;
+    video: globalThis.File | ElectronFile;
 }
 
 export interface FileWithCollection extends UploadAsset {
--- a/src/utils/upload/index.ts
+++ b/src/utils/upload/index.ts
@@ -1,4 +1,4 @@
-import { FileWithCollection, Metadata } from 'types/upload';
+import { ElectronFile, FileWithCollection, Metadata } from 'types/upload';
 import { EnteFile } from 'types/file';
 import { convertToHumanReadable } from 'utils/billing';
 import { formatDateTime } from 'utils/file';
@@ -74,6 +74,6 @@ export function getUploadLogs() {
         .map((log) => `[${formatDateTime(log.timestamp)}] ${log.logLine}`);
 }
 
-export function getFileNameSize(file: File) {
+export function getFileNameSize(file: File | ElectronFile) {
     return `${file.name}_${convertToHumanReadable(file.size)}`;
 }
--- a/src/services/readerService.ts
+++ b/src/services/readerService.ts
@@ -1,3 +1,5 @@
+import { ElectronFile } from 'types/upload';
+
 export async function getUint8ArrayView(
     reader: FileReader,
     file: Blob
@@ -37,6 +39,17 @@ export function getFileStream(
     const chunkCount = Math.ceil(file.size / chunkSize);
     return {
         stream,
+        chunkCount,
+    };
+}
+
+export async function getElectronFileStream(
+    file: ElectronFile,
+    chunkSize: number
+) {
+    const chunkCount = Math.ceil(file.size / chunkSize);
+    return {
+        stream: await file.createReadStream(),
         chunkCount,
     };
 }
--- a/src/services/upload/livePhotoService.ts
+++ b/src/services/upload/livePhotoService.ts
@@ -2,6 +2,7 @@ import { FILE_TYPE } from 'constants/file';
 import { LIVE_PHOTO_ASSET_SIZE_LIMIT } from 'constants/upload';
 import { encodeMotionPhoto } from 'services/motionPhotoService';
 import {
+    ElectronFile,
     FileTypeInfo,
     FileWithCollection,
     LivePhotoAssets,
@@ -23,7 +24,7 @@ interface LivePhotoIdentifier {
 }
 
 interface Asset {
-    file: File;
+    file: File | ElectronFile;
     metadata: Metadata;
     fileTypeInfo: FileTypeInfo;
 }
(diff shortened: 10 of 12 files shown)


rushitote:
After clicking the upload button the modal looks like this: (@vishnukvmd)

image

vishnukvmd:
@rushitote Can we have something like this?

image

vishnukvmd:
Here are SVGs for the icons

file_upload
folder_upload

rushitote:
@vishnukvmd  does this look fine?

image

vishnukvmd:
> @vishnukvmd  does this look fine?
> 
> 
> 
> image
> 
> 

Could you reduce the width by 25%?

rushitote:
image

vishnukvmd:
Could you also make sure that both options have the same height and each item within (icon, text, >) is vertically centered? Everything else LGTM.

rushitote:
The arrow icons were not properly centered, now it does look fine to me.

src/components/pages/gallery/Upload.tsx
```diff
@@ -121,11 +147,39 @@ export default function Upload(props: Props) {
         setProgressView(true);
     };
 
+    const uploadFailedFiles = async () => {
```
abhinavkgrd, line 150: I don't think we need this, What I was thinking was that we could just trigger a new upload by setting electronFiles State with the failed files na ?

src/components/pages/gallery/Upload.tsx
```diff
@@ -86,30 +92,50 @@ export default function Upload(props: Props) {
         );
     }, []);
 
+    useEffect(() => {
+        isElectron() &&
+            ImportService.getIfToUploadFilesExists().then((exists) => {
+                if (exists) {
+                    uploadFailedFiles();
```
abhinavkgrd, line 99: we can just call `setElectronFiles(getPendingUploads())` Where `getPendingUploads` would just return the electronFile for files in the pendingUpload list?

src/components/pages/gallery/Upload.tsx
```diff
@@ -245,10 +302,17 @@ export default function Upload(props: Props) {
             props.setUploadInProgress(true);
             props.closeCollectionSelector();
             await props.syncWithRemote(true, true);
+            // set state to not done
+            await ImportService.setToUploadFiles(
```
abhinavkgrd, line 306: so we are just setting and ToUploadFiles once at the start, and so if the upload fails midway, the files which were successfully uploaded would still be retired no? 

ideally, we should update this list after each successful upload.if this is not too difficult we should do this

src/services/importService.ts
```diff
@@ -0,0 +1,80 @@
...
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.showUploadDirsDialog();
+        }
+    }
+
+    async getToUploadFiles() {
+        if (this.allElectronAPIsExist) {
+            const { filesPaths, collectionName, collectionIDs } =
+                this.ElectronAPIs.getToUploadFiles();
+            const files = await getElectronFiles(filesPaths);
```
abhinavkgrd, line 49: I was thinking if we could just return ElectronFiles on `getToUploadFiles` and hence move this logic to the electron layer?

src/services/importService.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    async showUploadDirsDialog(): Promise<string[]> {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.showUploadDirsDialog();
+        }
+    }
+
+    async getToUploadFiles() {
+        if (this.allElectronAPIsExist) {
+            const { filesPaths, collectionName, collectionIDs } =
```
abhinavkgrd, line 47: if are just triggering a new upload for files, then we don't need to save the `collectionIDs` as those will be automatically retrieved as trying to create a Collection with existing name would return the existing Collection's ID

src/services/importService.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    async setToUploadFiles(
+        files: FileWithCollection[],
+        collections?: Collection[]
+    ) {
+        if (this.allElectronAPIsExist) {
+            let collectionName;
+            if (collections?.length > 0) {
+                collectionName = collections[0].name;
```
abhinavkgrd, line 65: we can just store the file paths and the  in case the user chose to create a new collection with a custom name we should store that  `collectionName` , everything else can be derived na

src/services/typeDetectionService.ts
```diff
@@ -29,8 +29,16 @@ export async function getFileType(
...
-        const typeResult = await extractFileType(reader, receivedFile);
-        const mimTypeParts = typeResult.mime?.split('/');
+        let mimTypeParts: string[];
+        let ext: string;
+        if (receivedFile instanceof File) {
+            const typeResult = await extractFileType(reader, receivedFile);
+            mimTypeParts = typeResult.mime?.split('/');
+            ext = typeResult.ext;
+        } else {
+            mimTypeParts = receivedFile.type.mimeType.split('/');
```
abhinavkgrd, line 39: we can use the same library `File-Type` to get File type from stream

src/services/upload/fileService.ts
```diff
@@ -121,7 +128,10 @@ export async function encryptFile(
     Get the original file name for edited file to associate it to original file's metadataJSON file 
     as edited file doesn't have their own metadata file
 */
-function getFileOriginalName(file: File) {
+function getFileOriginalName(file: File | ElectronFile) {
+    if (!(file instanceof File)) {
```
abhinavkgrd, line 132: we don't need to do this na, we are just accessing the name property which exists on both `File` and `ElectronFile`?

src/services/upload/metadataService.ts
```diff
@@ -26,11 +27,20 @@ const NULL_PARSED_METADATA_JSON: ParsedMetadataJSON = {
 };
 
 export async function extractMetadata(
-    receivedFile: File,
+    receivedFile: File | ElectronFile,
     fileTypeInfo: FileTypeInfo
 ) {
     let extractedMetadata: ParsedExtractedMetadata = NULL_EXTRACTED_METADATA;
     if (fileTypeInfo.fileType === FILE_TYPE.IMAGE) {
+        if (!(receivedFile instanceof File)) {
```
abhinavkgrd, line 35: I think `exifr` library can handle streams too, please check that once and if yes then we could pass a stream and avoid loading the whole file in memory

src/utils/upload/index.ts
```diff
@@ -30,8 +31,9 @@ export function areFilesSame(
 ): boolean {
     if (
         existingFile.fileType === newFile.fileType &&
-        existingFile.creationTime === newFile.creationTime &&
-        existingFile.modificationTime === newFile.modificationTime &&
+        Math.abs(existingFile.creationTime - newFile.creationTime) < 1e6 &&
+        Math.abs(existingFile.modificationTime - newFile.modificationTime) <
+            1e6 && // 1 second
```
abhinavkgrd, line 36: better to store it as a constant named `A_SEC_IN_MICROSECONDS`

src/services/importService.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    async showUploadDirsDialog(): Promise<string[]> {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.showUploadDirsDialog();
+        }
+    }
+
+    async getToUploadFiles() {
```
abhinavkgrd, line 45: `getPendingUploads`

src/components/pages/gallery/Upload.tsx
```diff
@@ -14,19 +14,21 @@ import { logError } from 'utils/sentry';
...
-import { FileWithCollection } from 'types/upload';
+import { ElectronFile, FileWithCollection } from 'types/upload';
 
 const FIRST_ALBUM_NAME = 'My First Album';
 
 interface Props {
     syncWithRemote: (force?: boolean, silent?: boolean) => Promise<void>;
     setBannerMessage: (message: string | JSX.Element) => void;
-    acceptedFiles: File[];
+    acceptedFiles: File[] | ElectronFile[];
```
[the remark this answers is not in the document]
rushitote: Yes, we don't need it now. It was added when I was testing something.

src/services/importService.ts
rushitote: I think we still need `collectionIDs`. Because we don't have the collection name if the user chose to upload to some existing collection as then this `collections` array would be empty. I couldn't find a way to find out the collection name except for storing the IDs. Should we have some state for `collectionName` that would get updated (or is there already something)?

src/services/upload/metadataService.ts
rushitote: I couldn't find any option for stream, but there is something to read exif data from a file path - should we add a function for this in Electron layer?

src/components/pages/gallery/Upload.tsx
rushitote: I don't think this is ideal, because wouldn't the user then will see the whole upload flow again?

src/services/importService.ts
rushitote: I have removed the part of storing `collectionIDs`, now it will store `collectionName` in both cases of an existing or a new collection.

src/components/pages/gallery/Upload.tsx
rushitote: I think we need the `collectionName` too.

src/components/pages/gallery/Upload.tsx
abhinavkgrd: we can have a special state as we have for the first upload, where we check if it's the first upload then we don't show the collection creation and `collectionSelector` modal and directly create and upload to new collection name "First Album", we can have something like that for this too na?

src/components/pages/gallery/Upload.tsx
abhinavkgrd: ya ya what I meant was we can avoid the two-step process of first getting the filepath and then calling  the getElectronFile Function to get the  ElectronFile instance,  instead we can directly send the `ElectronFile` instance from the electron layer na.

we can send this as function response
```
{
collectionName: string,
files:ElectronFile[]
}
```

src/components/pages/gallery/Upload.tsx
rushitote: Cool, will do that way.

src/services/upload/uploadManager.ts
```diff
@@ -256,6 +260,11 @@ class UploadManager {
     private async uploadMediaFiles(mediaFiles: FileWithCollection[]) {
         logUploadInfo(`uploadMediaFiles called`);
         this.filesToBeUploaded.push(...mediaFiles);
+
+        if (isElectron()) {
+            this.remainingFiles.push(...mediaFiles);
```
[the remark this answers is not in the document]
rushitote: `filesToBeUploaded` contains files which have not yet been pushed into the queue while `remainingFiles` contains files which have not yet been successfully uploaded/ignored. So, `fileToBeUploaded` was missing some files.

src/services/typeDetectionService.ts
```diff
@@ -53,6 +63,14 @@ async function extractFileType(reader: FileReader, file: File) {
     return getFileTypeFromBlob(reader, fileChunkBlob);
 }
 
+async function extractFileTypeFromElectronFile(file: ElectronFile) {
+    const stream = await file.stream();
+    const reader = stream.getReader();
+    const { value } = await reader.read();
+    const fileTypeResult = await FileType.fromBuffer(value);
```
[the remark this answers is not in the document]
rushitote: It needs `ReadStream`, and our `stream` function returns a `ReadableStream`.

src/services/importService.ts
```diff
@@ -0,0 +1,92 @@
...
+            return {
+                files,
+                collectionName,
+            };
+        }
+    }
+
+    async setToUploadFiles(
+        files: FileWithCollection[],
+        collections?: Collection[]
```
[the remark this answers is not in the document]
rushitote: because it was optional parameter in `uploadFiles` too.

src/pages/_app.tsx
```diff
@@ -501,6 +501,9 @@ const GlobalStyles = createGlobalStyle`
         width: 30vw;
     }
 
+    .manageLinkHeader:hover{
```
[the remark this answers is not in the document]
rushitote: it was added to fix a merge conflict

src/components/pages/gallery/Upload.tsx
```diff
@@ -322,6 +311,8 @@ export default function Upload(props: Props) {
             setProgressView(false);
             throw err;
         } finally {
+            // set state to done
+            await ImportService.setDoneUploadingFiles();
```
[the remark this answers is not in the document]
rushitote: I fear if there is some error and we don't mark it as done in the `finally` block, it will keep reappearing every time the app starts up. 😅
Should I move it to where it was earlier?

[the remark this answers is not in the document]
abhinavkgrd: ohh okk cool understood

src/services/importService.ts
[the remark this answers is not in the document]
abhinavkgrd: oh my bad 🤦 ... I will fix this,you can ignore this

src/services/typeDetectionService.ts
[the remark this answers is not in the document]
abhinavkgrd: ohh sad 😢

src/components/pages/gallery/Upload.tsx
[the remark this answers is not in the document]
abhinavkgrd: but then we would lose the records of the files that failed na? because when we set `done` we clear the `filePaths` array  ?

src/services/upload/uploadManager.ts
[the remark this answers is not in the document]
abhinavkgrd: oh no no, you are confused a bit, `filesToBeUploaded` is the upload queue, so it has the files whose upload has not started yet.

so it is in real the `pendingUploads`

src/components/pages/gallery/Upload.tsx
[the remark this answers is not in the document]
rushitote: Yeah right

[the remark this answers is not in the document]
abhinavkgrd: we are anyways getting rid of the `done` property so we can ignore this

src/services/upload/uploadManager.ts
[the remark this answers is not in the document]
abhinavkgrd: what we should do is have another list for failed uploads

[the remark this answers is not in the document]
rushitote: Yes, but it does not contain those files which are in the upload process, right? When I tried interrupting the app midway in upload, those files who were not fully uploaded got missed.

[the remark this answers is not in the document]
abhinavkgrd: ok understood, then what we should be doing is marking files as uploaded and removing them from the pendingUpload queue

src/services/typeDetectionService.ts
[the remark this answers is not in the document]
abhinavkgrd: still, we should not read the full file into memory, we type can be detected from a 4KB chunk of the file, is there some way we could just get that much file data?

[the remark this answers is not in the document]
rushitote: The `reader.read()` returns a max. `4 MB` chunk at a time as defined [here](https://github.com/ente-io/bhari-frame/pull/15/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R3).

rushitote:
There was an issue with the desktop app not detecting suggested collection name on Windows, fixed it by making the path unix style.

src/services/upload/uploadManager.ts
```diff
@@ -327,6 +331,18 @@ class UploadManager {
                 this.failedFiles.push(fileWithCollection);
             }
 
+            if (fileUploadResult && isElectron()) {
+                this.remainingFiles = this.remainingFiles.filter(
+                    (file) =>
+                        (file.file as ElectronFile).path !==
```
abhinavkgrd, line 337: lets create a helper function called `areSameElectronFile` to do this

src/services/upload/videoMetadataService.ts
```diff
@@ -1,9 +1,15 @@
 import { NULL_EXTRACTED_METADATA } from 'constants/upload';
 import ffmpegService from 'services/ffmpeg/ffmpegService';
+import { ElectronFile } from 'types/upload';
 import { logError } from 'utils/sentry';
 
-export async function getVideoMetadata(file: File) {
+export async function getVideoMetadata(file: File | ElectronFile) {
     let videoMetadata = NULL_EXTRACTED_METADATA;
+    if (!(file instanceof File)) {
```
abhinavkgrd, line 8: can we just pass `ElectronFile` as `File` to `getVideoMetadata` ?

src/services/importService.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    async hasPendingUploads(): Promise<boolean> {
+        if (this.allElectronAPIsExist) {
+            return this.ElectronAPIs.hasPendingUploads();
+        }
+    }
+
+    async getElectronFile(filePath: string): Promise<ElectronFile> {
```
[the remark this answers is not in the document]
abhinavkgrd: basically we should not be working with file paths to handle any logic of converting from the path to ElectronFile in` bhari-frame`

src/components/pages/gallery/Upload.tsx
```diff
@@ -121,11 +153,32 @@ export default function Upload(props: Props) {
...
+        }
+    };
+
     function analyseUploadFiles(): AnalysisResult {
-        if (props.acceptedFiles.length === 0) {
+        if (toUploadFiles.current.length === 0) {
             return null;
         }
-        const paths: string[] = props.acceptedFiles.map((file) => file['path']);
+        if (isElectron() && !isUploadDirs) {
```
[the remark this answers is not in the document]
rushitote: Should we do that? Like, I was trying to upload just a single file from my general Downloads folder, and it created a collection for that - should it be like this?

src/components/pages/gallery/Upload.tsx
```diff
@@ -258,6 +319,8 @@ export default function Upload(props: Props) {
             setProgressView(false);
             throw err;
         } finally {
+            // set state to done
+            await ImportService.setDoneUploadingFiles();
```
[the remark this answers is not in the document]
rushitote: Yep, removed the functions.

src/services/importService.ts
rushitote: Yep, it did worked.

src/services/upload/fileService.ts
```diff
@@ -40,7 +45,9 @@ export async function readFile(
     );
     logUploadInfo(`reading file datal${getFileNameSize(rawFile)} `);
     let filedata: Uint8Array | DataStream;
-    if (rawFile.size > MULTIPART_PART_SIZE) {
+    if (!(rawFile instanceof File)) {
+        filedata = await getElectronFileStream(rawFile, FILE_READER_CHUNK_SIZE);
```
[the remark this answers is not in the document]
rushitote: The electron files already always use stream to upload, why do we need this check?

src/services/upload/videoMetadataService.ts
rushitote: I am not sure if we can, since the ffmpeg function uses a blob as an argument (which the File type satisfies).

abhinavkgrd:
also we need to have some way to cancel the upload, once started on the desktop client 

refreshing was the usual way earlier but now it would auto-resume on refresh

src/components/pages/gallery/Upload.tsx
```diff
@@ -329,9 +402,16 @@ export default function Upload(props: Props) {
         }
     };
 
+    const cancelPendingDesktopUploads = () => {
+        UploadManager.clearRemainingFiles();
+        ImportService.updatePendingUploads([]);
+        setProgressView(false);
+        router.reload();
```
abhinavkgrd, line 409: we should call `syncWithRemote`, instead of reloading the whole page

src/components/pages/gallery/UploadProgress.tsx
```diff
@@ -286,6 +288,16 @@ export default function UploadProgress(props: Props) {
                     sectionInfo={constants.TOO_LARGE_INFO}
                 />
             </Modal.Body>
+            {props.isPendingDesktopUpload && (
```
abhinavkgrd, line 291: let's give this cancel button for web too if we are adding this

src/services/upload/uploadManager.ts
```diff
@@ -338,6 +352,10 @@ class UploadManager {
     async retryFailedFiles() {
         await this.queueFilesForUpload(this.failedFiles);
     }
+
+    clearRemainingFiles() {
+        this.remainingFiles = [];
```
abhinavkgrd, line 357: should we also clear `filesToBeUploaded` ?

src/services/upload/uploadManager.ts
```diff
@@ -327,20 +336,30 @@ class UploadManager {
...
+                    this.failedFilesWithResults.push({
+                        filePath: (fileWithCollection.file as ElectronFile)
+                            .path,
+                        fileUploadResult: FileUploadResults.BLOCKED,
+                    });
+                }
             }
 
-            if (fileUploadResult && isElectron()) {
+            if (fileUploadResult !== null && isElectron()) {
```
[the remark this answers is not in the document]
rushitote: Yeh since it's an enum it can have a `0` value. 😅

src/components/pages/gallery/Upload.tsx
```diff
@@ -261,6 +322,8 @@ export default function Upload(props: Props) {
             appContext.resetSharedFiles();
             props.setUploadInProgress(false);
             props.syncWithRemote();
+            if (isPendingDesktopUpload.current)
```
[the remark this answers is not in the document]
rushitote: It wouldn't have worked that way, since we wanted this value to be persisted till the upload is over as it is used in the cancel uploads button, but since we are now making the button visible for all uploads I think it doesn't matter now.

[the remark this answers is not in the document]
abhinavkgrd: cools then we can ignore this

src/utils/strings/englishConstants.tsx
```diff
@@ -684,13 +684,18 @@ const englishConstants = {
...
+    DISABLE_PASSWORD_MESSAGE:
+        'are you sure that you want to disable the password lock?',
     PASSWORD_LOCK: 'password lock',
     LOCK: 'lock',
     DOWNLOAD_UPLOAD_LOGS: 'debug logs',
     CHOOSE_UPLOAD_TYPE: 'Upload',
     UPLOAD_FILES: 'File Upload',
     UPLOAD_DIRS: 'Folder Upload',
+    NO_FAILED_UPLOADS: 'no failed uploads found',
+    FAILED_UPLOADS_HEADING: 'Failed Uploads',
```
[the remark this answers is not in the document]
rushitote: Will fix this in failed uploads PR, since this commit isn't in this one.

src/components/pages/gallery/Upload.tsx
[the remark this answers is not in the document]
rushitote: I have reverted back this change.

src/components/pages/gallery/Upload.tsx
```diff
@@ -121,11 +156,29 @@ export default function Upload(props: Props) {
...
+        }
+    };
+
     function analyseUploadFiles(): AnalysisResult {
-        if (props.acceptedFiles.length === 0) {
+        if (toUploadFiles.current.length === 0) {
             return null;
         }
-        const paths: string[] = props.acceptedFiles.map((file) => file['path']);
+        if (isElectron() && !isUploadDirs) {
```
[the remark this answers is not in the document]
rushitote: Can we just change the default value of `isUploadDirs` here to `true` ? We can make it to `true` at the end of every upload and it can become false if the user uploads via the file button, maybe can also give it a better name too. Would this work?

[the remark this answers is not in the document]
abhinavkgrd: I didn't understand how that will solve the problem 
To clarify-> I meant the case where the user drags and drops a file instead of using the upload button to upload the file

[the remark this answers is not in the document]
rushitote: Sorry, I was wrong - I assumed that drag and drop will have relative file paths like on the web and let that analyze function work as it does on web. But since Electron itself adds file paths to every file, it will lead to absolute file paths too.

src/services/importService.ts
```diff
@@ -0,0 +1,58 @@
...
+        */
+        if (collections.length === 1) {
+            collectionName = collections[0].name;
+        }
+        const filePaths = files.map((file) => (file.file as ElectronFile).path);
+        this.ElectronAPIs.setToUploadFiles(filePaths, collectionName);
+    }
+
+    updatePendingUploads(files: FileWithCollection[]) {
+        const filePaths = files.map((file) => (file.file as ElectronFile).path);
```
abhinavkgrd, line 54: this can also fail for `live photos`?

src/services/importService.ts
```diff
@@ -0,0 +1,58 @@
...
+                    to one album
+                hence saving the collection name when upload collection count is 1
+                helps the info of user choosing this options
+                and on next upload we can directly start uploading to this collection 
+        */
+        if (collections.length === 1) {
+            collectionName = collections[0].name;
+        }
+        const filePaths = files.map((file) => (file.file as ElectronFile).path);
+        this.ElectronAPIs.setToUploadFiles(filePaths, collectionName);
```
abhinavkgrd, line 50: I was thinking we can make a rename this as `setToUploadFilesAndCollection`

src/services/importService.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    updatePendingUploads(files: FileWithCollection[]) {
+        const filePaths = files.map((file) => (file.file as ElectronFile).path);
+        this.ElectronAPIs.updatePendingUploadsFilePaths(filePaths);
+    }
```
abhinavkgrd, line 56: we can use the same  `setToUploadFilesAndCollection` instead of having a separate function na ?

rushitote: No, then we will have to pass `collectionName` too right?

abhinavkgrd: we can pass null and have a check to update only its not `null` ?

or we can have separate functions for saving collections and files, the current system of a function to update "files and collection"  and another to update just files is a little weird. we can have lets say   `setToUploadFiles` and  `setToUploadCollection`

rushitote:
The stop uploads modal looks like this:

image

src/services/importService.ts
```diff
@@ -47,12 +47,23 @@ class ImportService {
...
+        this.ElectronAPIs.setToUploadFiles(filePaths);
+        this.ElectronAPIs.setToUploadCollection(collectionName);
     }
 
     updatePendingUploads(files: FileWithCollection[]) {
-        const filePaths = files.map((file) => (file.file as ElectronFile).path);
-        this.ElectronAPIs.updatePendingUploadsFilePaths(filePaths);
+        const filePaths = [];
+        for (const file of files) {
+            if (file.isLivePhoto) {
```
abhinavkgrd, line 57: let's call this `fileWithCollection`only , not to get confused with the actual file

src/utils/upload/index.ts
```diff
@@ -74,6 +77,13 @@ export function getUploadLogs() {
...
-export function getFileNameSize(file: File) {
+export function getFileNameSize(file: File | ElectronFile) {
     return `${file.name}_${convertBytesToHumanReadable(file.size)}`;
 }
+
+export function areFileWithCollectionsSame(
+    firstFile: FileWithCollection,
+    secondFile: FileWithCollection
+): boolean {
+    return firstFile.localID === secondFile.localID;
```
ua741, line 88: Sorry, i haven't read the complete code. 
What exactly is this `local_ID` and how does it get set? Say if I upload a desktop folder for X files. And then I delete few files from that folder and try to upload again, will the localID change?

src/utils/upload/index.ts
```diff
@@ -30,8 +31,10 @@ export function areFilesSame(
 ): boolean {
     if (
         existingFile.fileType === newFile.fileType &&
-        existingFile.creationTime === newFile.creationTime &&
-        existingFile.modificationTime === newFile.modificationTime &&
+        Math.abs(existingFile.creationTime - newFile.creationTime) <
+            A_SEC_IN_MICROSECONDS &&
```
ua741, line 35: Shall we document why we changing this logic and comparing them like double with a precision of `A_SEC_IN_MICROSECONDS` (why not 10 or 100 MILLISECONDS?).

src/types/upload/index.ts
```diff
@@ -68,14 +68,25 @@ export interface ProgressUpdater {
     setHasLivePhotos: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
+export interface ElectronFile {
```
ua741, line 71: nit: For future, It would be nice if we can add some document the need for creating this new `ElectronFile` type.

src/types/upload/index.ts
```diff
@@ -68,14 +68,25 @@ export interface ProgressUpdater {
...
+    name: string;
+    path: string;
```
ua741, line 73: Does the path contain full path? like `C:/Users/Rushi` or path within the folder that user pointed for upload.

src/services/upload/videoMetadataService.ts
ua741: super nit: `file instanceOf ElectronFile` instead of `!(file instanceof File)`.  This will also call out this line whenever someone will look for reference of `ElectronFile`

src/components/pages/gallery/Upload.tsx
```diff
@@ -84,32 +100,51 @@ export default function Upload(props: Props) {
             },
             props.setFiles
         );
+
+        if (isElectron()) {
+            ImportService.getPendingUploads().then(
+                ({ files, collectionName }) => {
```
ua741, line 106: super nit: `files` -> `electronFiles`. The term file is already overloaded with system file, enteFile and now electronFile :)

src/components/pages/gallery/Upload.tsx
```diff
@@ -121,11 +156,29 @@ export default function Upload(props: Props) {
...
+        collectionName: string
+    ) => {
+        try {
+            if (files && files?.length > 0) {
+                isPendingDesktopUpload.current = true;
+                pendingDesktopUploadCollectionName.current = collectionName;
+                props.setElectronFiles(files);
+            }
+        } catch (e) {
+            logError(e, 'Failed to get previously failed files');
```
ua741, line 170: nit: review and check if we need to update the error message? What we are doing inside method doesn't seem to be matching the the error copy. 
Also, I see that we are swallowing the error here. Always try to document the reason for such cases :). It will help future maintainer and they won't think it as a mistake.

src/utils/upload/index.ts
ua741: Also, do we nee to change it for all files or just electronFiles?

src/utils/upload/index.ts
abhinavkgrd: local_ID is just a sequential ID given to the files queued for upload, it's only used in the context of an upload session to refer to those files.

yes on a new upload, all the files will be assigned a new local_ID, but we don't persist the old_ids anywhere so it won't be a problem

src/utils/upload/index.ts
rushitote: The value was chosen such that it could be much higher than what any user would keep to resist fingerprinting ([context](https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified#reduced_time_precision)). I thought up to a second would be fine, but need to think what could be the smallest such value.

And yes we need it for all files since eventually the user will have a mix of uploads done by electronFiles and normal files and `existingFile` can be any of them.

src/types/upload/index.ts
rushitote: Yes, full path.

src/services/upload/videoMetadataService.ts
rushitote: The `instanceof` needs a constructor for comparison which the `File` is, but `ElectronFile` is currently just a interface type.

src/components/pages/gallery/Upload.tsx
rushitote: I removed the try-catch since the functionality had changed a bit and I don't think we need it now.

And yes will keep in mind to document any such things. 🙂

src/utils/upload/index.ts
ua741: Got it. One second seems okay. We should some of this context as documentation, so that it will help us in future while maintaining this logic.

abhinavkgrd (approved):

PR #453, opened 2022-03-23
Deduplicate

rushitote:
## Description

Added deduplication service and basic rendering for it.

## Test Plan
- Tested on various duplicate files and it did work similar to that on mobile.
- There is an issue with missing files - the server responds with all the duplicate files including which are in trash but when we check which of them to display we only select those which are in any albums other than trash and the rest are added to list of missing files. So, currently the web app logs an error containing missing files every time for this. This does not affect any functionality though. So, should we just skip adding the file to missing list if it's found in trash?


src/services/fileService.ts
```diff
@@ -157,15 +158,42 @@ export const trashFiles = async (filesToTrash: EnteFile[]) => {
...
-        });
+
+        for (const file of filesToTrash) {
+            trashBatch.items.push({
+                collectionID: file.collectionID,
+                fileID: file.id,
+            });
+            currentBatchSize++;
+            if (currentBatchSize >= MAX_TRASH_BATCH_SIZE) {
+                await HTTPService.post(
```
vishnukvmd, line 174: You could extract this into a separate function to avoid code duplication?

src/services/fileService.ts
```diff
@@ -157,15 +158,42 @@ export const trashFiles = async (filesToTrash: EnteFile[]) => {
...
-            'X-Auth-Token': token,
-        });
+
+        for (const file of filesToTrash) {
+            trashBatch.items.push({
+                collectionID: file.collectionID,
+                fileID: file.id,
+            });
+            currentBatchSize++;
+            if (currentBatchSize >= MAX_TRASH_BATCH_SIZE) {
```
vishnukvmd, line 173: Could compare with `trash.items.length` and discard `currentBatchSize`?

src/components/pages/deduplicate/SelectedDuplicatesOptions.tsx
```diff
@@ -0,0 +1,117 @@
...
+import DeleteIcon from 'components/icons/DeleteIcon';
+import { SetDialogMessage } from 'components/MessageDialog';
+import Navbar from 'components/Navbar';
+import React from 'react';
+import { OverlayTrigger } from 'react-bootstrap';
+import styled from 'styled-components';
+import constants from 'utils/strings/constants';
+
+interface Props {
+    setDialogMessage: SetDialogMessage;
```
abhinavkgrd, line 12: this complete component is a duplicate of `selectedFileOptions` na, 

I think a better way of doing this would be to have [this part](https://github.com/ente-io/bada-frame/pull/453/files#diff-1411c25bb4dd24de9bd932af0cbdb2bdb5ded3ba3c4bfba771c436c312c99b86R79-R112)  as a component and pass `isDeduplicatePage` to `selectedFileOptions` and conditionally render those options

src/pages/deduplicate/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+                    style={{
+                        position: 'absolute',
+                        top: '1em',
+                        left: '1em',
+                        zIndex: 10,
+                    }}
+                    onClick={() => {
+                        router.push(PAGES.GALLERY);
+                    }}>
+                    <LeftArrow />
```
abhinavkgrd, line 304: let wrap the svg with [`IconButton`](https://github.com/ente-io/bada-frame/blob/86a692f908d9e0bc240203a78a7db78cf43ad534/src/components/Container.ts#L21),  it will improve its styling

src/services/deduplicationService.ts
```diff
@@ -0,0 +1,133 @@
...
+    size: number;
+}
+
+class DeduplicationService {
+    public async getDuplicateFiles() {
+        try {
+            const dupes = await this.fetchDuplicateFileIDs();
+            let ids: number[] = [];
+            for (const dupe of dupes) {
+                ids = ids.concat(dupe.fileIDs);
```
abhinavkgrd, line 28: we can use `map` for this na?

src/pages/deduplicate/index.tsx
```diff
@@ -0,0 +1,309 @@
+import EnteSpinner from 'components/EnteSpinner';
```
abhinavkgrd, line 1: please refactor this whole page and only have whats absolutely necessary

src/services/deduplicationService.ts
```diff
@@ -0,0 +1,133 @@
...
+        size: number;
+    }>;
+}
+
+interface DuplicateFiles {
+    files: EnteFile[];
+    size: number;
+}
+
+class DeduplicationService {
```
abhinavkgrd, line 22: doesn't make sense to make it a class if we don't have any local state, we can just make it a collection of exported functions like `fileService`

src/services/deduplicationService.ts
abhinavkgrd: also we don't use it anywhere ?

src/services/deduplicationService.ts
```diff
@@ -0,0 +1,133 @@
...
+                    result.push({
+                        files,
+                        size: dupe.size,
+                    });
+                }
+            }
+
+            if (missingFileIDs.length > 0) {
+                logError(
+                    new Error(`Missing files: ${missingFileIDs}`),
```
abhinavkgrd, line 61: as we know these are files that are in the trash. so no point in logging them na? we can just ignore them?

src/services/deduplicationService.ts
```diff
@@ -0,0 +1,133 @@
...
+                );
+            }
+
+            return result;
+        } catch (e) {
+            logError(e, 'failed to get duplicate files');
+        }
+    }
+
+    public async clubDuplicatesByTime(dupes: DuplicateFiles[]) {
```
abhinavkgrd, line 72: this is different from what I expected.  Let's say I have five files, which have the same size, but they are such that 3 of them have one timestamp while the other 2 have a different timestamp.

what this would do is show only 3 files with the same timestamp. 

but what I expected was that it would split them into two groups and show them as separate rows


did you discuss this with @vishnukvmd bhaiya ?

src/services/fileService.ts
```diff
@@ -264,3 +275,9 @@ export const updatePublicMagicMetadata = async (files: EnteFile[]) => {
         })
     );
 };
+
+async function trashFilesFromServer(trashBatch: TrashRequest, token: any) {
+    await HTTPService.post(`${ENDPOINT}/files/trash`, trashBatch, null, {
```
abhinavkgrd, line 280: we should have a  try-catch block wrapper around this

src/services/deduplicationService.ts
rushitote: I implemented the service as it was in the mobile app (https://github.com/ente-io/frame/blob/master/lib/services/deduplication_service.dart#L89), not sure why it is that way though.

src/services/deduplicationService.ts
rushitote: yep my mistake, was using this earlier to test something and we don't need this now

src/pages/deduplicate/index.tsx
rushitote: I did minimize it to as much possible but I am not sure what else we can remove as all the current components seem necessary.

rushitote:
There was a bug due to which the collection selector modal used to reappear when navigating from deduplicate page to gallery as one of the `useEffect` was triggered, it was fixed by keeping the `Upload` component on both pages. 

And another one was when there is no `fileSize` of some item in the deduplicate page (for a split second) when it is fetching, it caused an undefined variable error. So, made it 0 if it is undefined.

abhinavkgrd (dismissed):

abhinavkgrd (approved):

abhinavkgrd:
image

abhinavkgrd (approved):

PR #464, opened 2022-04-06
Upload Google Takeout

rushitote:
## Description

- Support for uploading Google zip takeouts directly without loading them entirely into the memory 
- This branch is based off of the `recover-failed-imports` branch since it reuses the `ImportService` and `ElectronFile` type

## Test Plan
- Tried uploading single and multiple zip takeouts


src/components/FailedUploadsModal.tsx
```diff
@@ -0,0 +1,118 @@
+import { FileUploadResults } from 'constants/upload';
```
abhinavkgrd, line 1: let's create a PR for failed upload option

abhinavkgrd (approved):

PR #493, opened 2022-04-25
fix can't read property of undefined (reading metadata) error  

abhinavkgrd:
## Description

It randomly occurred in the dev environment, unable to reproduce it consistently, but this should help avoid it from happening in any case

checking that the  `file` is not undefined or null before proceeding into the function will prevent any error caused by trying to access a property on an `undefined` variable 

## Test Plan

theoretically it should work


abhinavkgrd:
### Update

- change the logic to get files using `fileID` instead of index while updating their URLs
- added better logs

ua741 (approved):

PR #521, opened 2022-05-12
Worker uploads

rushitote:


src/utils/common/apiUtil.ts
```diff
@@ -51,3 +51,10 @@ export const getFamilyPortalURL = () => {
     }
     return `https://family.ente.io`;
 };
+
+export const getUploadURL = () => {
```
abhinavkgrd, line 55: lets call this `getUploadEndpoint`

PR #527, opened 2022-05-16
better handle upload error 

abhinavkgrd:
## Description

handle post upload task error gracefully  

## Test Plan

- tested that breaking errors like subscription expired and session expired are propagated 
- but other errors are caught and logged and don't break other uploads

### DIFF
--- a/src/services/upload/uploadManager.ts
+++ b/src/services/upload/uploadManager.ts
@@ -80,12 +80,6 @@ class UploadManager {
         );
     }
 
-    async retryFailedFiles() {
-        await this.queueFilesForUpload(this.failedFiles, [
-            ...this.collections.values(),
-        ]);
-    }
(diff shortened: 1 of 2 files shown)


src/services/upload/uploadManager.ts
```diff
@@ -323,22 +339,10 @@ class UploadManager {
                     .get(file.collectionID)
                     .push(file);
             }
-            if (fileUploadResult === FileUploadResults.FAILED) {
-                this.failedFiles.push(fileWithCollection);
-                setData(LS_KEYS.FAILED_UPLOADS, {
```
ua741, line 328: Just to confirm, we don't need to track such failures now? What was the use-case for adding this in the first place?

ua741 (approved):

PR #537, opened 2022-05-18
File Hashing

rushitote:
## Description

Currently storing hashes for images and videos in `hash` and for live photos in `imageHash` and `videoHash`.

## Test Plan

- Tested locally uploading images / live photos / videos to check if hashes are generating
- Checked that the hash is same as on mobile
- Checked that uploading file with same hash and type as an existing file in that collection gets ignored
- Checked that uploading file with same hash, type and name as an existing file in other collections is not reuploaded but added to collection
- Checked that uploading file with same hash, type but different name as an existing file in other collections is reuploaded


abhinavkgrd:
@rushitote  please update the base branch `file-hash-dev` with master and resolve the merge conflicts

rushitote:
@abhinavkgrd done

src/utils/upload/index.ts
```diff
@@ -49,6 +76,33 @@ export function areFilesSame(
     }
 }
 
+export function fileHashExists(file: Metadata) {
+    if (file.hash || (file.imageHash && file.videoHash)) {
```
abhinavkgrd, line 80: we can directly return this expression as its a boolean expression 
i.e., 
`return file.hash || (file.imageHash && file.videoHash)`

src/utils/upload/index.ts
```diff
@@ -35,6 +58,10 @@ export function areFilesSame(
      * precision of file times to prevent timing attacks and fingerprinting.
      * Context: https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified#reduced_time_precision
      */
+    if (fileHashExists(existingFile) && fileHashExists(newFile)) {
+        return areFilesWithFileHashSame(existingFile, newFile);
```
abhinavkgrd, line 62: we should move this if block above the comment as that comment is related to the  second if block and not this one

src/utils/upload/index.ts
```diff
@@ -35,6 +58,10 @@ export function areFilesSame(
...
+    }
+
     if (
         existingFile.fileType === newFile.fileType &&
         Math.abs(existingFile.creationTime - newFile.creationTime) <
```
abhinavkgrd, line 67: we can also remove this if else block and directly return the expression

abhinavkgrd: after doing that we can make this function as 
```
if(has file hash){
 match file hash and type
}else{
    fallback to the older checks
}```

File: src/utils/upload/index.ts [hunk 4]
```diff
@@ -49,6 +76,33 @@ export function areFilesSame(
...
+export function areFilesWithFileHashSame(
+    existingFile: Metadata,
+    newFile: Metadata
+): boolean {
+    if (existingFile.fileType !== newFile.fileType) {
+        return false;
+    }
+    if (
+        existingFile.fileType === FILE_TYPE.IMAGE ||
+        existingFile.fileType === FILE_TYPE.VIDEO
```
abhinavkgrd, line 95: can we avoid having two condition by checking if file is livePhoto instead

src/utils/upload/index.ts
```diff
@@ -49,6 +76,33 @@ export function areFilesSame(
...
+        return true;
+    }
+    return false;
+}
+
+export function areFilesWithFileHashSame(
+    existingFile: Metadata,
+    newFile: Metadata
+): boolean {
+    if (existingFile.fileType !== newFile.fileType) {
```
abhinavkgrd, line 90: we also need to check if file names are same no ?

src/utils/upload/index.ts
```diff
@@ -20,6 +21,28 @@ export function fileAlreadyInCollection(
...
+    collectionID: number
+) {
+    if (!fileHashExists(newFileMetadata)) {
+        return null;
+    }
+
+    for (const existingFile of existingFiles) {
+        if (
+            existingFile.collectionID !== collectionID &&
+            fileHashExists(existingFile.metadata) &&
```
abhinavkgrd, line 36: we are doing the same check here as `areFilesWithFileHashSame` , aren't we ?

src/services/upload/uploader.ts
```diff
@@ -53,6 +56,26 @@ export default async function uploader(
...
+            collection.id
+        );
+
+        if (sameFileInOtherCollection) {
+            logUploadInfo(
+                `same file in other collection found for  ${fileNameSize}`
+            );
+            const resultFile = Object.assign({}, sameFileInOtherCollection);
+            resultFile.collectionID = collection.id;
+            await addToCollection(collection, [resultFile]);
```
abhinavkgrd, line 71: it would be better if we return the encrypted new file from addToCollection and return that forward, we won't need the new skipDecryption then

src/utils/upload/index.ts
rushitote: Ah yes, sorry

abhinavkgrd (approved):

PR #538, opened 2022-05-18
Added file hashing

rushitote:
## Description

#537

### DIFF
--- a/src/services/upload/livePhotoService.ts
+++ b/src/services/upload/livePhotoService.ts
@@ -47,11 +47,17 @@ export function getLivePhotoFileType(
     };
 }
 
-export function getLivePhotoMetadata(imageMetadata: Metadata) {
+export function getLivePhotoMetadata(
+    imageMetadata: Metadata,
+    videoMetadata: Metadata
+) {
     return {
         ...imageMetadata,
         title: getLivePhotoName(imageMetadata.title),
         fileType: FILE_TYPE.LIVE_PHOTO,
+        imageHash: imageMetadata.hash,
+        videoHash: videoMetadata.hash,
+        hash: undefined,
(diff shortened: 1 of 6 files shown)


src/services/upload/livePhotoService.ts
```diff
@@ -47,11 +47,17 @@ export function getLivePhotoFileType(
...
+    imageMetadata: Metadata,
+    videoMetadata: Metadata
+) {
     return {
         ...imageMetadata,
         title: getLivePhotoName(imageMetadata.title),
         fileType: FILE_TYPE.LIVE_PHOTO,
+        imageHash: imageMetadata.hash,
+        videoHash: videoMetadata.hash,
+        hash: undefined,
```
abhinavkgrd, line 60: don't need this

abhinavkgrd: we should make hash as optional

src/utils/upload/index.ts
```diff
@@ -93,3 +94,10 @@ export function areFileWithCollectionsSame(
 ): boolean {
     return firstFile.localID === secondFile.localID;
 }
+
+export async function getFileHash(file: File | ElectronFile) {
+    const hash = await cryptoGenericHash(
+        new Uint8Array(await file.arrayBuffer())
```
abhinavkgrd, line 100: let's keep this in crypto util

src/services/upload/livePhotoService.ts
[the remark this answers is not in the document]
rushitote: No, since we are destructuring `imageMetadata` there will be a `hash` present.

abhinavkgrd (approved):

PR #544, opened 2022-05-19
Use getFileStream instead of file.stream

rushitote:
## Test Plan

- Tested it generates same hash.

### DIFF
--- a/src/utils/crypto/index.ts
+++ b/src/utils/crypto/index.ts
@@ -7,8 +7,10 @@ import { getActualKey, getToken } from 'utils/common/key';
 import { setRecoveryKey } from 'services/userService';
 import { logError } from 'utils/sentry';
 import { ComlinkWorker } from 'utils/comlink';
-import { ElectronFile } from 'types/upload';
+import { DataStream, ElectronFile } from 'types/upload';
 import { cryptoGenericHash } from './libsodium';
+import { getElectronFileStream, getFileStream } from 'services/readerService';
+import { ENCRYPTION_CHUNK_SIZE } from 'constants/crypto';
 
 export interface B64EncryptionResult {
     encryptedData: string;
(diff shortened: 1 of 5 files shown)


src/utils/crypto/index.ts
```diff
@@ -200,7 +202,12 @@ export async function encryptWithRecoveryKey(key: string) {
 export default CryptoWorker;
 
 export async function getFileHash(file: File | ElectronFile) {
-    const stream = await file.stream();
-    const hash = await cryptoGenericHash(stream);
+    let filedata: DataStream;
+    if (file instanceof File) {
+        filedata = getFileStream(new FileReader(), file, ENCRYPTION_CHUNK_SIZE);
```
abhinavkgrd, line 207: so using `ENCRYPTION_CHUNK_SIZE` feels weird when we  are not dealing with encryption, so have  `FILE_READER_CHUNK_SIZE` when referring to the chunk generated by the file reader, internally it's the same

src/utils/crypto/index.ts
```diff
@@ -200,7 +202,16 @@ export async function encryptWithRecoveryKey(key: string) {
 export default CryptoWorker;
 
 export async function getFileHash(file: File | ElectronFile) {
-    const stream = await file.stream();
-    const hash = await cryptoGenericHash(stream);
+    let filedata: DataStream;
+    if (file instanceof File) {
+        filedata = getFileStream(
+            new FileReader(),
```
abhinavkgrd, line 208: creating a new FileReader every time can cause memory issues, it would be better if we create a instance of file reader in `extractMetadataFromFiles`  and pass that for every file

abhinavkgrd (approved):

abhinavkgrd (approved):

PR #545, opened 2022-05-19
Dedupe using file hash

rushitote:
## Test Plan

- Tested that only files with same file hashes are being grouped and shown

### DIFF
--- a/src/services/deduplicationService.ts
+++ b/src/services/deduplicationService.ts
@@ -2,7 +2,7 @@ import { EnteFile } from 'types/file';
 import { getEndpoint } from 'utils/common/apiUtil';
 import { getToken } from 'utils/common/key';
 import { logError } from 'utils/sentry';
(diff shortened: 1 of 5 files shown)


rushitote:
Looks like this:

image

src/services/deduplicationService.ts
```diff
@@ -113,6 +116,64 @@ export function clubDuplicatesByTime(dupes: DuplicateFiles[]) {
...
+
+        const filteredFiles = dupe.files.filter((file) => {
+            return hasFileHash(file.metadata);
+        });
+
+        if (filteredFiles.length <= 1) {
+            continue;
+        }
+
+        const dupesSortedByFileHash = filteredFiles.map((file) => {
```
abhinavkgrd, line 133: calling this `dupesSortedByFileHash` is a little weird as we have not sorted them  till now, lets's chain the sort call with this if we want to do so

abhinavkgrd (approved):

PR #558, opened 2022-05-23
Make club by file hashes default in dedupe

rushitote:
## Description

Looks like this:

image

## Test Plan

- Tested locally by manipulating dedupe options.

### DIFF
--- a/src/components/pages/gallery/SelectedFileOptions/DeduplicateOptions.tsx
+++ b/src/components/pages/gallery/SelectedFileOptions/DeduplicateOptions.tsx
@@ -76,15 +76,13 @@ export default function DeduplicateOptions({
                     width: '1em',
                     height: '1em',
                 }}
-                value={
-                    deduplicateContext.clubSameFileHashesOnly ? 'true' : 'false'
-                }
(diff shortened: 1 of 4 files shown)


src/components/pages/gallery/SelectedFileOptions/DeduplicateOptions.tsx
```diff
@@ -76,15 +76,13 @@ export default function DeduplicateOptions({
...
                     height: '1em',
                 }}
-                value={
-                    deduplicateContext.clubSameFileHashesOnly ? 'true' : 'false'
-                }
+                value={deduplicateContext.clubBySameSize ? 'true' : 'false'}
                 onChange={() => {
-                    deduplicateContext.setClubSameFileHashesOnly(
-                        !deduplicateContext.clubSameFileHashesOnly
+                    deduplicateContext.setClubBySameSize(
```
abhinavkgrd, line 81: why do we need this?

rushitote: I replaced state for hashes by state for same size since now hashes is default. Should we instead keep the hashes state?

PR #592, opened 2022-06-12
better password key derivation failed messsage

abhinavkgrd:

### DIFF
--- a/src/utils/strings/englishConstants.tsx
+++ b/src/utils/strings/englishConstants.tsx
@@ -749,7 +749,7 @@ const englishConstants = {
         </>
     ),
     WEAK_DEVICE:
-        "The device you're using is not powerful enough for the encryption of your photos. Please try to log in on desktop, or download the ente mobile app.",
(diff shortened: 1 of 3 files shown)


src/utils/strings/englishConstants.tsx
```diff
@@ -748,6 +748,8 @@ const englishConstants = {
             </p>
         </>
     ),
+    WEAK_DEVICE:
+        "The device you're using is not powerful enough for the encryption of your photos. Please try to log in on desktop, or download the ente mobile app.",
```
mnvr, line 752: Thanks for remembering this change!

I feel we should go with the message I'd suggested though. This message is shorter but it doesn't say the same thing. The problem is not the device (because if the problem was the device the it won't work on the app either). The problem is the mobile browser.

mnvr: "The mobile web browser you're using is not powerful enough for the encryption of your photos. Please try to log in to ente on your computer, or download the ente mobile app."

mnvr (approved):

PR #601, opened 2022-06-16
better handle metadata extraction error 

abhinavkgrd:
### problem
files whose metadata extraction failed didn't get a map entry and caused a `can't read property of undefined` error when was they were accessed

### solution

putting null values, for such cases to avoid the error


ua741 (approved):

ua741:
Can you please test the change first before landing this?

abhinavkgrd:
tested by manually throwing an error during from metadata extraction function

PR #604, opened 2022-06-18
Added syncing ML generated object tags with server

rushitote:
## Description

^

## Test Plan

- Tested locally that previous files and newly uploaded files are getting their tags updated.


abhinavkgrd:
cleaning up the repo, deleting PR, will keep the branch around for later

PR #615, opened 2022-06-24
fix complete upload crash- on localstrorage write fail

abhinavkgrd:
## Description

this is to fix the Neeraj bhaiya issue of upload crashing with a generic upload message

so, we don't let the single file error bubble up, we just log and add it to the progress bar view

## Test Plan

tested locally


ua741:
Here, will the user see on the UI which file failed to upload?

abhinavkgrd:
> 

yep, we will show the file in the `failed` section

ua741 (approved):

PR #672, opened 2022-07-27
show exif location as fallback if metadata is missing location 

### LINKED ISSUE #661: Map feature did not work with some coordinates
Hello,

on some photos, the "Show Map" feature is not working.
I send you two screens, one where you see the missing link and one where you see the full EXIF data with GPS coordinates:
image
image

abhinavkgrd:
## Description

so for some reason, some files don't have location metadata but the EXIF data has the location value, maybe the file was uploaded when we had an EXIF parsing bug.

For now, have added a fallback to the EXIF location if that's available 

TODO FOR FUTURE:  backfill the location from EXIF if it's missing in enteFile metadata

fixes #661

## Test Plan

tested locally

### DIFF
--- a/src/components/PhotoSwipe/InfoDialog/index.tsx
+++ b/src/components/PhotoSwipe/InfoDialog/index.tsx
@@ -47,15 +47,13 @@ export function FileInfo({
     const [location, setLocation] = useState<Location>(null);
 
     useEffect(() => {
-        if (
-            !location &&
-            metadata &&
-            (metadata.longitude || metadata.longitude === 0)
-        ) {
-            setLocation({
-                latitude: metadata.latitude,
-                longitude: metadata.longitude,
-            });
+        if (!location && metadata) {
+            if (metadata.longitude || metadata.longitude === 0) {
+                setLocation({
+                    latitude: metadata.latitude,
+                    longitude: metadata.longitude,
+                });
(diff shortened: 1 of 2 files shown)


src/components/PhotoSwipe/InfoDialog/index.tsx
```diff
@@ -28,8 +42,30 @@ export function FileInfo({
...
     exif,
     scheduleUpdate,
-}) {
+}: Iprops) {
     const appContext = useContext(AppContext);
+    const [location, setLocation] = useState<Location>(null);
+
+    useEffect(() => {
+        if (!location && metadata) {
+            if (metadata.longitude || metadata.longitude === 0) {
```
rushitote, line 51: This check should be like this, right? Also, we will check for latitude too, right?
```
metadata.longitude && metadata.longitude !== 0
```

abhinavkgrd: so, `longitude = 0` is a valid value but `if(metadata.longitude) ` will return `false` for it, so added an `or` condition to allow that.

>Also, we will check for latitude too, right?

if either both of the coordinates will be present or neither will be present, so just checking for one will do that job

rushitote (approved):

PR #676, opened 2022-07-29
Backfill missing location metadata

abhinavkgrd:
## Description
added   `editedLongitude` and `editedLatitude` in `pubMetadata` and use it to set location from exif

## Test Plan
- [ ] test with a file uploaded intentionally without location data and then checking if exif data is getting sent as edited coordinates

### DIFF
--- a/src/components/PhotoSwipe/InfoDialog/index.tsx
+++ b/src/components/PhotoSwipe/InfoDialog/index.tsx
@@ -58,29 +58,32 @@ export function FileInfo({
     }, [metadata]);
 
     useEffect(() => {
-        const main = async () => {
-            if (!location && exif) {
-                const exifLocation = getEXIFLocation(exif);
-                if (exifLocation.latitude || exifLocation.latitude === 0) {
-                    setLocation(exifLocation);
-                    try {
-                        let updatedFile = await updateFileLocation(
-                            currentItem,
-                            exifLocation
-                        );
-                        updatedFile = (
-                            await updateFilePublicMagicMetadata([updatedFile])
-                        )[0];
-                        updateExistingFilePubMetadata(currentItem, updatedFile);
-                        scheduleUpdate();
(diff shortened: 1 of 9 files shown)


src/components/PhotoSwipe/InfoDialog/index.tsx
```diff
@@ -58,12 +58,28 @@ export function FileInfo({
...
+                        )[0];
+                        updateExistingFilePubMetadata(currentItem, updatedFile);
+                        scheduleUpdate();
+                    } catch (e) {
+                        logError(e, 'failed to update file location');
+                    }
+                }
             }
-        }
+        };
```
ua741, line 81: can we define this method outside `useEffect` and call it here? Also document the reason why we are doing this.. May be method name can be descriptive enough.. Something like `_updateFileMetadataWithExifLocationData`

src/components/PhotoSwipe/InfoDialog/index.tsx
```diff
@@ -58,12 +58,28 @@ export function FileInfo({
...
-            const exifLocation = getEXIFLocation(exif);
-            if (exifLocation.latitude || exifLocation.latitude === 0) {
-                setLocation(exifLocation);
+        const main = async () => {
+            if (!location && exif) {
+                const exifLocation = getEXIFLocation(exif);
+                if (exifLocation.latitude || exifLocation.latitude === 0) {
+                    setLocation(exifLocation);
+                    try {
+                        let updatedFile = await updateFileLocation(
```
ua741, line 67: Should we also verify the ownership of the file? On Mobile, we let user view metadata for shared files.

abhinavkgrd: Oh ya, nice catch! I have added the check to verify it is part of shared collection and prevent update call in that case

PR #680, opened 2022-08-08
refactor conversion for preview

abhinavkgrd:
## Description

refactored convertForPreview to be used only for generating preview URLs of file 

and separate out the conversion of live photo and heic image into separate functions.

## Test Plan

tested locally for both normal gallery and public-URL gallery and tested live photo preview

### DIFF
--- a/src/services/publicCollectionDownloadManager.ts
+++ b/src/services/publicCollectionDownloadManager.ts
@@ -5,7 +5,7 @@ import {
 import CryptoWorker from 'utils/crypto';
 import {
     generateStreamFromArrayBuffer,
-    convertForPreview,
+    getRenderableFileURL,
     createTypedObjectURL,
 } from 'utils/file';
 import HTTPService from './HTTPService';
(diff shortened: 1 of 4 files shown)


src/services/publicCollectionDownloadManager.ts
```diff
@@ -113,15 +117,15 @@ class PublicCollectionDownloadManager {
...
-                    const convertedBlobs = await convertForPreview(
-                        file,
-                        fileBlob
-                    );
-                    return convertedBlobs.map((blob) =>
-                        URL.createObjectURL(blob)
-                    );
+                    return await getRenderableFileURL(file, fileBlob);
+                } else {
+                    return [
```
vishnukvmd, line 122: Let's wrap this in an object instead of returning an array?

abhinavkgrd: That will touch many more code paths than I want to change right now.

Currently, I refactored the function to perform a single task, which was the main issue.

using object vs array is more of a design choice. 

Planning to  do it in a separate PR later on

vishnukvmd (approved):

PR #693, opened 2022-08-30
Refactor upload

abhinavkgrd:
## Description

- moved stuff from uploader  component to util and types and constants 
- cleaner post upload action
- few variable rename 
- minor refactor like created pre and postUploadAction in uploader Component
- better duplicate file upload handling
- fixes the issue where the upload type selector does not close even after the upload starts
- fixes issue -> previous folder upload state was used even when the user was trying to upload files 

## Test Plan

- [x] tried single file upload
- [x] tried folder upload 
- [x] check if previous upload state files coming in the new upload 
- [x] tried duplicate file upload in different collection
- [x] tried duplicate file in the same collection
- [x] check analysis result and getCollectionWiseFiles 
- [x] tested multiple uploads without a refresh to check if the state was properly cleared
- [x] fix choice selector closing on upload start
- [x] block send drop/ upload when one already running running
- [x] check both electron and normal flow
- [x] check logic moved to pre and post-upload action in uploader Component
- [x] check to add symlink upload flow
- [x] check all upload result cases once 
- [x]  FAILED
- [x] BLOCKED
- [x] ADDED_SYMLINK
- [x] UPLOADED
- [x] UPLOADED_WITH_STATIC_THUMBNAIL
- [x] ALREADY_UPLOADED
- [x] tested upload shared collection files to the self album
- [x] tested personal files only used for depuping  
- [x] tested retry


src/utils/upload/index.ts
```diff
@@ -120,3 +131,60 @@ export function areFileWithCollectionsSame(
...
+                commonPathPrefix.lastIndexOf('/') + 1
+            );
+        }
+    }
+    return {
+        suggestedCollectionName: commonPathPrefix || null,
+        multipleFolders: firstFileFolder !== lastFileFolder,
+    };
+}
+export function getCollectionWiseFiles(toUploadFiles: File[] | ElectronFile[]) {
```
ua741, line 172: rename to getCollectionToFilesMap or some thing similar? Also, if not a big change, let's define return type too?

Edit: Before reading the code, I thought it was mapping collectionID to files. But that's definitely not the case. 
Let's rename it so that it's easy to guess what the method is doing without reading the code.

src/utils/upload/index.ts
```diff
@@ -120,3 +131,60 @@ export function areFileWithCollectionsSame(
...
+    const collectionWiseFiles = new Map<string, (File | ElectronFile)[]>();
+    for (const file of toUploadFiles) {
+        const filePath = file['path'] as string;
```
ua741, line 175: file['path'] will always be non-null/empty? Even when user has dragged few files right after sign up?

src/utils/upload/index.ts
```diff
@@ -120,3 +131,60 @@ export function areFileWithCollectionsSame(
...
+
+        let folderPath = filePath.substring(0, filePath.lastIndexOf('/'));
+        if (folderPath.endsWith(METADATA_FOLDER_NAME)) {
```
ua741, line 178: Didn't follow what are we doing here. Should we add a comment explaining what's happening here and may be why too if necessary?

src/utils/upload/index.ts
```diff
@@ -120,3 +131,60 @@ export function areFileWithCollectionsSame(
 ): boolean {
     return firstFile.localID === secondFile.localID;
 }
+
+export function analyseUploadFiles(
```
ua741, line 135: What's the purpose of this method? And when/why does is return `NULL_ANALYSIS_RESULT` without doing any anlysis?

src/services/upload/uploader.ts
```diff
@@ -50,40 +46,27 @@ export default async function uploader(
...
-            );
-            const resultFile = Object.assign({}, sameFileInOtherCollection);
-            resultFile.collectionID = collection.id;
-            await addToCollection(collection, [resultFile]);
-            return {
-                fileUploadResult: UPLOAD_RESULT.UPLOADED,
-                uploadedFile: resultFile,
-                skipDecryption: true,
-            };
+        const existingFile = findMatchingExistingFile(existingFiles, metadata);
```
ua741, line 49: there could be multiple files which might match the hash.. Here, we are only returning first match. Is that by design? Or will it cause issue?

src/utils/upload/index.ts
abhinavkgrd: This function analyses the upload files, suggests the name of the LCA of the files as the collectionName if they are to be uploaded in one collection.

And also returns if the file there are nested folder, in the folder the user is trying to upload.

So for `UPLOAD_TYPE.FILES` it returns NULL_ANALYSIS_RESULT because
1. The browser return path with respect to the LCA ,so for just files,they would have no common ancestor hence we can't suggest a collection name  
2. and there will be no nested folder, as they are just files

src/utils/upload/index.ts
abhinavkgrd: yes,it will always be `not-null/empty` this is the path will respect to the root folder the user is trying to upload.

If user is just uploading bunch of files with then it will be the fileName

src/services/upload/uploader.ts
abhinavkgrd: ahh...there is a case where the file is in multiple collection, one the current upload collection (say `x` ) and one another collection (say `y`).

we might get `y` collection's file and we will try to add a symlink for it in collection `x` and that will return a no-op from the server , and we would show the file as uploaded, instead of showing it as ignored already in collection.

src/utils/upload/index.ts
```diff
@@ -169,22 +169,42 @@ export function analyseUploadFiles(
...
+//            / |  \
+//           b  j   c
+//          /|\    /  \
+//         e f g   h  i
+//
+// The files will grouped into 3 collections.
+// [a => [j],
+// b => [e,f,g],
+// c => [h, i]]
+export function groupFilesBasedOnParentFolder(
```
abhinavkgrd, line 185: check and throw error if folder is null

src/utils/upload/index.ts
```diff
@@ -169,22 +169,42 @@ export function analyseUploadFiles(
...
+    toUploadFiles: File[] | ElectronFile[]
+) {
+    const collectionNameToFilesMap = new Map<string, (File | ElectronFile)[]>();
     for (const file of toUploadFiles) {
         const filePath = file['path'] as string;
 
         let folderPath = filePath.substring(0, filePath.lastIndexOf('/'));
+        // If the parent folder of a file is "metadata"
+        // we consider it to be part of the parent folder
```
abhinavkgrd, line 194: check case where user has collection named "metadata"

src/utils/upload/index.ts
abhinavkgrd: rename METADATA_FOLDER_NAME to ENTE_METADAT_FOLDER

abhinavkgrd: verify if all files are json

src/services/upload/uploader.ts
abhinavkgrd: check for ownership and trashed

src/components/Upload/Uploader.tsx
```diff
@@ -107,11 +93,13 @@ export default function Uploader(props: Props) {
     const pendingDesktopUploadCollectionName = useRef<string>('');
     const uploadType = useRef<UPLOAD_TYPE>(null);
```
abhinavkgrd, line 94: mention only set for upload type for file/folder picker upload

src/services/upload/uploader.ts
abhinavkgrd: done this

src/utils/upload/index.ts
abhinavkgrd: didn't do the verify JSON files 😅  ...was a big change

ua741 (approved):

ua741 (approved):

PR #696, opened 2022-09-05
Account delete

abhinavkgrd:
## Description

Free users can now initiate self-delete

## screenshots

image

the yes feedback button opens the user mail client

on no delete account  press 

for free accounts (having API response (`allowDelete: true` )

image

and for paid user 

image

on the delete option press 

image

which has handled authentication failure 

image

on authentication success, the challenge is solved and we make the delete account the account call and then logout the user 

## Test Plan

checked API call and authentication failure cases 
and  checked the happy path once


src/utils/strings/englishConstants.tsx
```diff
@@ -774,6 +774,31 @@ const englishConstants = {
     WEAK_DEVICE:
         "The web browser you're using is not powerful enough to encrypt your photos. Please try to log in to ente on your computer, or download the ente mobile/desktop app.",
     DRAG_AND_DROP_HINT: 'Or drag and drop into the ente window',
+    ASK_FOR_FEEDBACK: (
+        <>
+            <p>We'll be sorry to see you go. Are you facing some issue?</p>
+            <p>
+                Please write to us at{' '}
+                <Link href="mailto:account-deletion@ente.io">
```
ua741, line 782: feedback email id?

ua741:
Please test and confirm that there's no regression in any flow which might be touched during refactor:
 -> Sign up 
 -> Recovery Password
 -> Change Password

ua741 (dismissed):

ua741 (approved):

PR #714, opened 2022-09-17
Add batching to delete from trash

abhinavkgrd:
## Description

add batching to delete from trash 

## Test Plan

tested locally

### DIFF
--- a/src/types/file/index.ts
+++ b/src/types/file/index.ts
@@ -58,7 +58,3 @@ export interface TrashRequestItems {
     fileID: number;
     collectionID: number;
 }
-
-export interface DeleteFilesRequest {
-    fileIDs: number[];
-}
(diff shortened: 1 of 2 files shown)


src/types/file/index.ts
```diff
@@ -58,3 +58,7 @@ export interface TrashRequestItems {
     fileID: number;
     collectionID: number;
 }
+
+export interface DeleteFilesRequest {
```
ua741, line 62: Do we need this extra interface, if you are just internally using it in the `FileService`?

ua741 (approved):

src/types/file/index.ts
abhinavkgrd: no, not really that necessary , as the `request data`  is simply a array

abhinavkgrd: added it as the request body was passed from between functions

ua741 (approved):

PR #717, opened 2022-09-22
fix handling of null data saved in local-storage

abhinavkgrd:
## Description

add better `getData`  null | undefined data check 

## Test Plan

tested locally


ua741 (approved):

abhinavkgrd:
some issue with new login 😬 , will fix it and then release

PR #718, opened 2022-09-23
Fix error logging

abhinavkgrd:
## Description

- move non-private sentry config to sentry.properties
- make the default environment to `production`, so logging error to sentry is the default behaviour if no flags are set
- and logging to console is special behaviour which happens when an environment is explicitly set to  `development`

## Test Plan

tested on preview deployment error logging is working


abhinavkgrd:
closing because of misleading branch name

PR #719, opened 2022-09-23
Store sentry config in code

abhinavkgrd:
## Description

- move non-private sentry config to sentry.properties
- make the default environment to `production`, so logging error to sentry is the default behaviour if no flags are set
- and logging to console is special behaviour which happens when an environment is explicitly set to  `development`

## Test Plan

tested on preview deployment error logging is working

### DIFF
--- a/src/constants/sentry/index.ts
+++ b/src/constants/sentry/index.ts
@@ -1,10 +1,15 @@
+export const ENV_DEVELOPMENT = 'development';
+export const ENV_PRODUCTION = 'production';
+
 export const getSentryDSN = () =>
     process.env.NEXT_PUBLIC_SENTRY_DSN ??
     'https://60abb33b597c42f6a3fb27cd82c55101@sentry.ente.io/2';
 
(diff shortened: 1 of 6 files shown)


src/constants/sentry/index.ts
```diff
@@ -1,10 +1,15 @@
+export const ENV_DEVELOPMENT = 'development';
+export const ENV_PRODUCTION = 'production';
+
 export const getSentryDSN = () =>
     process.env.NEXT_PUBLIC_SENTRY_DSN ??
     'https://60abb33b597c42f6a3fb27cd82c55101@sentry.ente.io/2';
 
 export const getSentryENV = () =>
-    process.env.NEXT_PUBLIC_SENTRY_ENV ?? 'development';
+    process.env.NEXT_PUBLIC_SENTRY_ENV ?? ENV_PRODUCTION;
```
abhinavkgrd, line 9: intentional

vishnukvmd (approved):

PR #730, opened 2022-10-10
disable location search

abhinavkgrd:
## Description

don't get location suggestions during the search 

## Test Plan

tested locally, 
 no API calls for location results are made


mnvr (approved):
🙏

PR #731, opened 2022-10-10
Fix scrollbar color

abhinavkgrd:
## Description

use the MUI colour scheme setting option to change web default components to use dark theme colour 
 https://mui.com/material-ui/react-css-baseline/#color-scheme 
https://web.dev/color-scheme/

image

## Test Plan

tested with browser stack on windows (chrome and firefox ) and mac

### DIFF
--- a/src/components/WatchFolder/styledComponents.tsx
+++ b/src/components/WatchFolder/styledComponents.tsx
@@ -3,14 +3,11 @@ import { Box } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import VerticallyCentered from 'components/Container';
 
-export const MappingsContainer = styled(Box)(({ theme }) => ({
+export const MappingsContainer = styled(Box)(() => ({
     height: '278px',
     overflow: 'auto',
     '&::-webkit-scrollbar': {
         width: '4px',
-    },
-    '&::-webkit-scrollbar-thumb': {
-        backgroundColor: theme.palette.secondary.main,
     },
 }));
 
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -239,7 +239,7 @@ export default function App({ Component, err }) {
             </Head>
 
             <ThemeProvider theme={darkThemeOptions}>
-                <CssBaseline />
+                <CssBaseline enableColorScheme />
                 {showNavbar && <AppNavbar />}
                 <MessageContainer>
                     {offline && constants.OFFLINE_MSG}
--- a/src/themes/darkThemeOptions.tsx
+++ b/src/themes/darkThemeOptions.tsx
@@ -82,6 +82,9 @@ const darkThemeOptions = createTheme({
                 body: {
                     fontFamily: ['Inter', 'sans-serif'].join(','),
                     letterSpacing: '-0.011em',
+                    '::-webkit-scrollbar-thumb': {
+                        backgroundColor: 'rgba(256, 256, 256, 0.12)',
+                    },
                 },
                 strong: { fontWeight: 900 },
             },
--- a/src/themes/darkThemeOptions.tsx
+++ b/src/themes/darkThemeOptions.tsx
@@ -82,9 +82,6 @@ const darkThemeOptions = createTheme({
                 body: {
                     fontFamily: ['Inter', 'sans-serif'].join(','),
                     letterSpacing: '-0.011em',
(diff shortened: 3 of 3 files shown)


src/components/WatchFolder/styledComponents.tsx
```diff
@@ -3,15 +3,12 @@ import { Box } from '@mui/material';
...
 
-export const MappingsContainer = styled(Box)(({ theme }) => ({
+export const MappingsContainer = styled(Box)(() => ({
     height: '278px',
     overflow: 'auto',
     '&::-webkit-scrollbar': {
         width: '4px',
     },
-    '&::-webkit-scrollbar-thumb': {
-        backgroundColor: theme.palette.secondary.main,
```
mnvr, line 13: I'm not sure what other components MUI refers to when it says in the docs:

> switching between "light" and "dark" modes of native components such as scrollbar, using the color-scheme CSS property

What was the issue here -- was this "webkit-scrollbar-thumb" override not working in all browsers?

abhinavkgrd: >What was the issue here -- was this "webkit-scrollbar-thumb" override not working in all browsers? 
No, this was no longer required, as the global enableColorScheme setting will handle the colour for this scrollbar too, so I removed it. 

> I'm not sure what other components MUI refers to when it says in the docs:
this article mention https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme)
>This includes form controls, scrollbars, and the used values of CSS system colours.

and https://developer.mozilla.org/en-US/docs/Web/CSS/system-color mentions the system colours

mnvr: I understand that the CSS color scheme, e.g. we'll use it when we do light mode. But here we're giving up control of the colors to use for the scroll bar. So these will not be colors as per our theme colors, these will be colors that MUI has picked.

This might be fine for scrollbars, if the scrollbar colors are slightly different from our rest of the theme colors. But what was bothering me a bit that what are the other "system controls" that MUI refers to when they say they will take control of the colors of "native components such as scrollbar".

If it is easy to find out,  let us find out. Otherwise we can merge this, and just remember that we turned on this flag (in case we find some part of the UI that's not changing as per our theme).

mnvr: > Any we don't set explicitly will follow the dark theme?

My issue is that I'm not sure of what this setting is doing. For me, on macOS, the scrollbar works fine both on master and your branch, in both light and dark modes. If we tell MUI to start styling it, we then take a dependency on MUI colors. But maybe these are the "system colors" as you said, and it would be a good default.

In any case, the scrollbar is not the issue. In the first example for the MDN page, you can see it is also changing the background color of text fields. We might care about these things at some point.

Also, after you enable this, the same problem that happens with the title bar will also happen with the scroll bar. Rest of the app is in dark mode (because we don't support it yet), only the scroll bar would be in light mode.

abhinavkgrd: > Also, after you enable this, the same problem that happens with the title bar will also happen with the scroll bar. Rest of the app is in dark mode (because we don't support it yet), only the scroll bar would be in light mode.

no no, that won't happen as the theme to use is taken from the option we have set in mui options

> If enableColorScheme is provided to CssBaseline, native components color will be set by applying [color-scheme](https://web.dev/color-scheme/) on <html>. The value used is provided by the theme property theme.palette.mode.

abhinavkgrd: In the first example for the MDN page, you can see it is also changing the background color of text fields. We might care about these things at some point.

yes, I agree, but when we would care about it, we would have set it explicitly.

till then the default dark themed color for it is better than a white themed color ?

mnvr: Cool!

mnvr (approved):

src/components/WatchFolder/styledComponents.tsx
mnvr: > > In the first example for the MDN page, you can see it is also changing the background color of text fields. We might care about these things at some point.
> 
> yes, I agree, but when we would care about it, we would have set it explicitly.
> 
> till then the default dark themed color for it is better than a white themed color ?

Yes that's true, so let's merge this for now. We can see if it causes any mismatches when we do the light mode them.

PR #747, opened 2022-10-21
Prevent yarn timeout 

abhinavkgrd:
## Description

prevent network timeout during the building of bada-frame under bhari-frame 

image

this comment suggested that increasing yarn timeout helps

https://github.com/mui/material-ui/issues/12432#issuecomment-411046157

## Test Plan

pending


.yarnrc
```diff
@@ -0,0 +1 @@
+network-timeout 500000
```
abhinavkgrd, line 1: this is the only change

ua741 (approved):

PR #776, opened 2022-11-19
Export caption in the sidecar file

ua741:
## Test Plan
YOLO. Will test before next release.

### DIFF
--- a/src/utils/export/index.ts
+++ b/src/utils/export/index.ts
@@ -128,15 +128,18 @@ export const dedupe = (files: any[]) => {
 
 export const getGoogleLikeMetadataFile = (
     fileSaveName: string,
-    metadata: Metadata
+    file: EnteFile
 ) => {
+    const metadata: Metadata = file.metadata;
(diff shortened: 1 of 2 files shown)


src/utils/export/index.ts
```diff
@@ -128,15 +128,18 @@ export const dedupe = (files: any[]) => {
...
+    const metadata: Metadata = file.metadata;
     const creationTime = Math.floor(metadata.creationTime / 1000000);
     const modificationTime = Math.floor(
         (metadata.modificationTime ?? metadata.creationTime) / 1000000
     );
+    const captionValue: string = file?.pubMagicMetadata?.data?.caption ?? '';
     return JSON.stringify(
         {
             title: fileSaveName,
+            caption: captionValue,
```
abhinavkgrd, line 142: should we skip the key if the caption is empty ?

abhinavkgrd (approved):

abhinavkgrd (approved):

PR #779, opened 2022-11-21
Sort upload files

abhinavkgrd:
## Description

sort upload files 

## Test Plan

tested locally

### DIFF
--- a/src/services/upload/uploadManager.ts
+++ b/src/services/upload/uploadManager.ts
@@ -103,6 +103,12 @@ class UploadManager {
         collections: Collection[]
     ) {
         try {
+            filesWithCollectionToUploadIn.sort(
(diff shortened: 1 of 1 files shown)


src/services/upload/uploadManager.ts
```diff
@@ -103,6 +103,12 @@ class UploadManager {
         collections: Collection[]
     ) {
         try {
+            filesWithCollectionToUploadIn.sort(
+                (aFileWithCollection, bFileWithCollection) =>
+                    (aFileWithCollection.file['path'] as string).localeCompare(
```
ua741, line 108: Are we confident that the file['path] will be always non-null/undefined? Not sure what will be the behaviour when path is missing.

ua741 (approved):

PR #808, opened 2022-12-04
don't report thumbnail upto size of 150KB

abhinavkgrd:
## Description
getting sentry log for thumbnail just greater than 100KB( like 100.58 KB)

updated the report large thumbnails only on condition that they are greater than 150KB  

image


abhinavkgrd:
@ua741, I just increased it to 150KB randomly, let me know if we want to keep it something lower

ua741 (approved):

PR #855, opened 2023-01-10
configure api util for staging

abhinavkgrd:
## Description

configured staging deployment connect to the staging server.

So that jishnu can test flows on staging environment.

## Test Plan

- tested that the normal and public uploads and  normal file/thumbnail and public-file/public-thumbnail download are working properly.  

- confirmed that non-staging deployment connect to production server

### DIFF
--- a/src/utils/common/apiUtil.ts
+++ b/src/utils/common/apiUtil.ts
@@ -84,9 +84,14 @@ It's a dev deployment (and should use the environment override for endpoints ) i
 2. when the URL opened is that of the staging album app, or
 3. if the app is running locally (hence node_env is development)
 */
-const isDevDeployment = () =>
-    runningInBrowser() &&
-    (process.env.NEXT_PUBLIC_ENTE_WEB_ENDPOINT === window.location.origin ||
-        process.env.NEXT_PUBLIC_ENTE_ALBUM_ENDPOINT ===
-            window.location.origin ||
-        process.env.NODE_ENV === 'development');
+const isDevDeployment = () => {
+    if (runningInBrowser()) {
+        return (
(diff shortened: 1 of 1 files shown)


src/utils/common/apiUtil.ts
```diff
@@ -50,23 +67,20 @@ export const getAlbumsURL = () => {
...
 };
 
 const isDevDeployment = () =>
-    process.env.NEXT_PUBLIC_ENTE_WEB_ENDPOINT === window.location.origin ||
-    process.env.NEXT_PUBLIC_ENTE_ALBUM_ENDPOINT === window.location.origin;
+    runningInBrowser() &&
+    (process.env.NEXT_PUBLIC_ENTE_WEB_ENDPOINT === window.location.origin ||
+        process.env.NEXT_PUBLIC_ENTE_ALBUM_ENDPOINT ===
+            window.location.origin ||
+        process.env.NODE_ENV === 'development');
```
ua741, line 86: should this be an and condition for process.env.NODE_ENV === 'development')

ua741: If not, why is that? May be we can add some documentation for this method

abhinavkgrd: Yes, it is as it should be.

So the condition says that it's a dev deployment (and should use the environment override for ENDPOINTS ) in three cases 
when 
1. when the app opened is the staging web app, or
2. when the app opened is the staging album app, or
3. if the app is running locally (hence node_env is `development`)

ua741 (approved):

ua741 (approved):

PR #871, opened 2023-01-17
Album icon change

Ananddubey01:
## Description

1. Changed archived Album icon color to fixed.strokeMutedWhite
2. moved album icon to left bottom
3. show share (people icon) for self shared albums
4. show link icon for albums with just public link

## Test Plan

Screenshot from 2023-01-18 13-19-20

Share Icon is for only those albums not been shared by anyone via email. Once you share any link shared albums via email, the link icon will disappear and people icon would be shown instead.

### DIFF
--- a/src/constants/collection/index.ts
+++ b/src/constants/collection/index.ts
@@ -16,6 +16,7 @@ export enum CollectionSummaryType {
     trash = 'trash',
     all = 'all',
     shared = 'shared',
+    sharedViaLink = 'sharedViaLink',
     archived = 'archived',
 }
 export enum COLLECTION_SORT_BY {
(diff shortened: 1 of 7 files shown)


src/constants/collection/index.ts
```diff
@@ -16,6 +16,7 @@ export enum CollectionSummaryType {
     trash = 'trash',
     all = 'all',
     shared = 'shared',
+    sharedViaLink = 'sharedViaLink',
```
abhinavkgrd, line 19: `sharedOnlyViaLink` name would be more clear (for what we are doing) and avoid any  confusion 

refer comment on `isSharedViaLink` for the reason

abhinavkgrd (changes requested):

abhinavkgrd (approved):

abhinavkgrd (approved):

PR #882, opened 2023-01-20
Album header change

Ananddubey01:
## Description
- [ ] Album Quick Options 
      - Download quick option is for all albums except 'all' and 'trash' albums
      - For Favorites, there is only the 'Download favorites' option.
      - For Trash, there is only an 'Empty trash' option
      - sharing icon/option is for all albums except 'all', 'trash', and 'favorites', however, it has different tooltips depending on share state
      

- [ ] Reordered and renamed Album options
- [ ] Updated Album Memory count size And icon positioning and size
- [ ] Added shared Album icon
## Test Plan
Screenshot from 2023-01-20 15-07-50
Screenshot from 2023-01-20 15-08-15
Screenshot from 2023-01-20 15-08-42
Screenshot from 2023-01-20 15-08-50
Screenshot from 2023-01-20 15-08-56
Screenshot from 2023-01-20 15-09-56


src/components/Collections/CollectionInfo.tsx
```diff
@@ -11,13 +11,25 @@ interface Iprops {
 export function CollectionInfo({ name, fileCount, endIcon }: Iprops) {
     return (
         <div>
+            <Typography variant="h3">{name}</Typography>
```
abhinavkgrd, line 14: I have forgotten, why there is that extra div, can we just put everything inside the FlexWrapper and remove the extra div

src/components/Collections/CollectionOptions/QuickOptions.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    </IconButton>
+                </Tooltip>
+            )}
+            {!(collectionSummaryType === CollectionSummaryType.trash) && (
+                <Tooltip
+                    title={
+                        collectionSummaryType ===
+                        CollectionSummaryType.favorites
+                            ? constants.DOWNLOAD_FAVOURITES
+                            : constants.CONFIRM_DOWNLOAD_COLLECTION
```
abhinavkgrd, line 55: constants.CONFIRM_DOWNLOAD_COLLECTION
 
 can you rename this to variable to DOWNLOAD_COLLECTION,
 
 thank you !!

src/components/Collections/CollectionOptions/index.tsx
```diff
@@ -232,30 +234,38 @@ const CollectionOptions = (props: CollectionOptionsProps) => {
...
-              CollectionSummaryType.incomingShare ? (
-                <SharedCollectionOption
-                    handleCollectionAction={handleCollectionAction}
-                />
-            ) : (
-                <AlbumCollectionOption
-                    IsArchived={IsArchived(activeCollection)}
-                    handleCollectionAction={handleCollectionAction}
-                />
+        <Box sx={{ display: 'inline-flex', gap: '16px' }}>
```
abhinavkgrd, line 237: you can use `FlexWrapper`  , instead of creating a Box

src/components/Collections/CollectionInfo.tsx
Ananddubey01: FlexWrapper will wrap 'Album name' and 'album memory count' in the same line which we don't want
Screenshot from 2023-01-20 16-58-35

src/components/Collections/CollectionOptions/index.tsx
Ananddubey01: `FlexWrapper`'s "width: 100%" property is causing the header to shrink.
`HorizontalFlex` is working fine though.

src/components/Collections/CollectionInfo.tsx
abhinavkgrd: ohh okk okk

abhinavkgrd (approved):

PR #894, opened 2023-01-25
fix file undefined error 

abhinavkgrd:
## Description

add proper file and not checks.

## Test Plan

testing locally


src/components/PhotoViewer/index.tsx
```diff
@@ -308,6 +308,7 @@ function PhotoViewer(props: Iprops) {
         });
         photoSwipe.listen('beforeChange', () => {
             const currItem = photoSwipe?.currItem as EnteFile;
+            if (!currItem) return;
```
ua741, line 311: if we do if(!photoSwipe?currItem), will it have same effect? If yes, should we add this as first statement inside the listener?

ua741 (approved):

ua741 (approved):

PR #901, opened 2023-01-30
fix sync for reshared album 

abhinavkgrd:
## Description

clear last sync time of a collection after it has been deleted 

## Test Plan

tested locally


src/services/collectionService.ts
```diff
@@ -197,6 +205,8 @@ export const syncCollections = async () => {
         if (!collection.isDeleted) {
             collections.push(collection);
             updationTime = Math.max(updationTime, collection.updationTime);
+        } else {
+            setCollectionLastSyncTime(collection, 0);
```
ua741, line 209: why not remove?

ua741 (approved):

PR #906, opened 2023-01-31
Uncategorised

Ananddubey01:
## Description

- [x] add a sidebar section like trash and archive
- [x] Uncategorised section options (only download option)
- [x] updated Remove from Album confirmation
- [x] file selection option
      These options will be avialable on uncategorized
      - move to album
      - trash file
   
- [x]  collection selector uncategoried not visible 

- [x]  update delete collection API
- [x]  updated Delete Album Confirmation

## Test Plan
Screenshot from 2023-02-02 22-36-25
Screenshot from 2023-01-31 21-23-41
Screenshot from 2023-02-02 22-38-14
Screenshot from 2023-01-31 21-24-38


abhinavkgrd:
- The uncategorized section is showing "download favorites" tooltip
image

- The download message should also say "download favorite"/ "download uncategorized"
image

in design
image

- the second option should be delete photos, also the "all" word is of primary color (`#fff`) and not `bold` 
image

image

abhinavkgrd:
the new design uses primary for the "yes remove" option

image

- The section shows zero when it doesn contain any item in the designs

image

abhinavkgrd:
If the uncategorized collection doesn't have any entry then no name is shown when I click on the sidebar section

https://user-images.githubusercontent.com/46242073/215959275-7bb12da9-4c94-4714-9412-c1e534b9fd77.mov

src/components/Sidebar/ShortcutSection.tsx
```diff
@@ -16,6 +18,21 @@ export default function ShortcutSection({
     collectionSummaries,
 }: Iprops) {
     const galleryContext = useContext(GalleryContext);
+    const [unCategorizedCollectionId, setUncategorezedId] = useState<number>();
+    useEffect(() => {
+        const main = async () => {
+            const unCategorisedCollection = await getUncategorizedCollection();
```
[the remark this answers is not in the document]
abhinavkgrd: we will be creating an uncategorized collection for the user if one doesn't exist, but the collection state is not updated so the UI won't show it 

Lets fix this by calling setCollection here 

you can update the galleryContext to have setCollection passed through it which we can use to update the state

src/constants/collection/index.ts
```diff
@@ -56,6 +62,7 @@ export const UPLOAD_NOT_ALLOWED_COLLECTION_TYPES = new Set([
     CollectionSummaryType.outgoingShare,
     CollectionSummaryType.sharedOnlyViaLink,
     CollectionSummaryType.trash,
+    CollectionSummaryType.uncategorized,
```
abhinavkgrd, line 65: uploading to uncategorized is allowed, as it is an actual collection

src/constants/collection/index.ts
```diff
@@ -41,12 +45,14 @@ export const COLLECTION_SORT_ORDER = new Map([
...
     [CollectionSummaryType.archive, 3],
     [CollectionSummaryType.trash, 4],
+    [CollectionSummaryType.uncategorized, 4],
 ]);
 
 export const SYSTEM_COLLECTION_TYPES = new Set([
     CollectionSummaryType.all,
     CollectionSummaryType.archive,
     CollectionSummaryType.trash,
+    CollectionSummaryType.uncategorized,
```
abhinavkgrd, line 55: so, as we discussed, `all`, `archive` and `trash` are just filters, not actuall collections, so they are marked as system collection

but uncategorized is an actual collection and hence it should not be treated as system collection

src/constants/collection/index.ts
abhinavkgrd: also, pls merge the main once, have removed `outgoingShare` and `sharedOnlyViaLink` , as `outgoingShare` and `sharedOnlyViaLink` are  user owned collections, which user should be able to upload to.

only incomingShare should be in this list as, they are someone else's collection in which user can't upload

src/pages/gallery/index.tsx
```diff
@@ -777,6 +778,10 @@ export default function Gallery() {
                                 activeCollection,
                                 collections
                             )}
+                            isUncategorizedCollection={isUncategorizedCollection(
```
abhinavkgrd, line 781: we should user collectionSummary to determine is an album is uncategorized.

That way we wouldn't have to loop through all collections

abhinavkgrd: also can you please update the above isFavoriteCollection to do so.

src/services/collectionService.ts
```diff
@@ -494,32 +497,42 @@ export const removeFromCollection = async (
     files: EnteFile[]
```
abhinavkgrd, line 497: we can improve this in few ways

- first the caller can send the list of existing collection. So for places where this function called multiple times. we would only get the collection collections once, similarly will calling get uncategorizedCollection we can pass it the list of collection, so that it doesn't have to fetch them from localStorage
- also in the current logic we would call move to collection `files.length` number of times, we can improve this by moving group files to one common collection if its possible

src/pages/gallery/index.tsx
abhinavkgrd: also there is isSharedCollection, can you also update

abhinavkgrd (changes requested):

src/components/Collections/CollectionOptions/index.tsx
```diff
@@ -138,7 +143,21 @@ const CollectionOptions = (props: CollectionOptionsProps) => {
...
+    const deleteButkeepFiles = async () => {
+        const allFiles = await getLocalFiles();
+        const collectionFiles = allFiles.filter((file) => {
+            return file.collectionID === activeCollection.id;
+        });
+        await CollectionAPI.removeFromCollection(
+            activeCollection.id,
+            collectionFiles
+        );
+        console.log(collectionFiles);
```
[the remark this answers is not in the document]
abhinavkgrd: so, let's pass the files and collection as galleryContext value and use it here.

instead of reading them from local db

src/services/collectionService.ts
```diff
@@ -532,6 +532,13 @@ export const deleteCollection = async (
     keepFiles: boolean
 ) => {
     try {
+        if (keepFiles) {
+            const allFiles = await getLocalFiles();
+            const collectionFiles = allFiles.filter((file) => {
+                return file.collectionID === collectionID;
+            });
+            await removeFromCollection(collectionID, collectionFiles);
```
abhinavkgrd, line 566: A small change, as we will anyways be needing`allFiles` in  `removeFromCollection` let's pass that from here

abhinavkgrd: The `removeFromCollection` can then be like

```
export const removeFromCollection = async (
    collectionID: number,
    toRemoveFiles: EnteFile[],
    allFiles?: EnteFile[]
) => {
    try {
        if (!allFiles) {
            allFiles = await getLocalFiles();
        }
        ...
        
 ```
 
 if the caller haven't sent the files then we don't need to read local files

File: src/components/Collections/CollectionOptions/QuickOptions.tsx [hunk 8]
```diff
@@ -25,6 +25,7 @@ export function QuickOptions({
             {!(
                 collectionSummaryType === CollectionSummaryType.trash ||
```
abhinavkgrd, line 26: not actually related to this but we should ideally write this conditional statements as

```
collectionSummaryType === CollectionSummaryType.trash ?
<TrashQuickOptions>:
collectionSummaryType === CollectionSummaryType.uncategoried?
<UncategoriedQuickOptions>:
...
```

That way the code will be more cleaner

src/utils/collection/index.ts
```diff
@@ -77,7 +77,7 @@ export function getSelectedCollection(
     return collections.find((collection) => collection.id === collectionID);
 }
 
-export function isSharedCollection(
+export function isIncomingSharedCollection(
```
abhinavkgrd, line 80: we can remove the three `isXYZCollection` function now ?as it is not used any where ?

src/types/collection/index.ts
```diff
@@ -37,7 +37,7 @@ export interface Collection
     > {
     key: string;
     name: string;
-    isSharedCollection?: boolean;
+    isIncomingSharedCollection?: boolean;
```
abhinavkgrd, line 40: this collection property is not used anywhere we can remove this too

abhinavkgrd (approved):

PR #916, opened 2023-02-04
Fix chrome possible out of memory warning

abhinavkgrd:
## Description
- rafactored photoframe 
- prevent updation of files state and did all updates on the photoFrame component
- simplified `mergeMetadata`, `sortFiles` 
- removed preservePhotoSwipeProps

## Test Plan
- [x] check `filteredData` and `files` changes are handled properly
- [x] confirm outdated udpateURL check is working
- [x] confirm `updateURL` is never called after the src URL is set
- [x] test `getLatestVersionFiles`
- [x] test `mergeMetadata`
- [x] test `sortFiles` 
- [x] test syncFiles -> try to break with error in middle of syncs
- [x] confirm "getThumbnail function works properly
        - preview card working properly,
        - check scrolling through the gallery doesn't cause re-render
        - sync doesn't cause renderers
        - range select and normal select works
        - it updates the file properly
        - in range select and select on click works

### DIFF
--- a/src/utils/photoFrame/index.ts
+++ b/src/utils/photoFrame/index.ts
@@ -60,6 +60,8 @@ export function updateFileMsrcProps(file: EnteFile, url: string) {
             `;
     } else if (file.metadata.fileType === FILE_TYPE.IMAGE) {
         file.src = url;
+    } else {
+        throw Error('Invalid file type');
     }
 }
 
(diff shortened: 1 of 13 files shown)


src/utils/photoFrame/index.ts
```diff
@@ -32,3 +37,106 @@ export async function pauseVideo(livePhotoVideo, livePhotoImage) {
...
+                <div class="pswp-item-container">
+                    <img src="${url}" onContextMenu="return false;"/>
+                    <div class="spinner-border text-light" role="status">
+                        <span class="sr-only">Loading...</span>
+                    </div>
+                </div>
+            `;
+    } else if (file.metadata.fileType === FILE_TYPE.IMAGE) {
+        file.src = url;
+    }
```
ua741, line 65: Add error for un-handled fileType

ua741 (approved):

PR #932, opened 2023-02-08
Quick options refactor

Ananddubey01:
## Description
Changed QuickOptions from a file to a collection of quickoption files. 
## Test Plan
Screenshot from 2023-02-08 09-53-13
Screenshot from 2023-02-08 09-53-18
Screenshot from 2023-02-08 09-53-31
Screenshot from 2023-02-08 09-53-54


src/components/Collections/CollectionOptions/QuickOptions/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+                <MoveToTrashOption
+                    handleCollectionAction={handleCollectionAction}
+                />
+            )}
+            {showDownloadQuickOption(collectionSummaryType) && (
+                <DownloadOption
+                    handleCollectionAction={handleCollectionAction}
+                    tooltipTitle={
+                        collectionSummaryType ===
+                        CollectionSummaryType.favorites
```
abhinavkgrd, line 38: can we move this logic to inside the `DownloadOption` component? we can pass it to the `collectionSummaryType`. can it can do this logic and show the appropriate tooltip

src/components/Collections/CollectionOptions/QuickOptions/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+}: Iprops) {
+    return (
+        <FlexWrapper sx={{ gap: '16px' }}>
+            {showTrashQuickOption(collectionSummaryType) && (
+                <MoveToTrashOption
+                    handleCollectionAction={handleCollectionAction}
+                />
+            )}
+            {showDownloadQuickOption(collectionSummaryType) && (
+                <DownloadOption
```
abhinavkgrd, line 34: should we name it `DownloadQuickOption`?

src/components/Collections/CollectionOptions/QuickOptions/MoveToTrashOption.tsx
```diff
@@ -0,0 +1,26 @@
...
+import constants from 'utils/strings/constants';
+
+interface Iprops {
+    handleCollectionAction: (
+        action: CollectionActions,
+        loader?: boolean
+    ) => (...args: any[]) => Promise<void>;
+}
+
+export function MoveToTrashOption({ handleCollectionAction }: Iprops) {
```
abhinavkgrd, line 14: Let's name it `EmptyTrashQuickOption` ?  as it is doing that... `emptyTrash` and `moveToTrash` are different things with different logic/ API ... so this naming can cause confusion

src/components/Collections/CollectionOptions/QuickOptions/ShareOption.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface Iprops {
+    handleCollectionAction: (
+        action: CollectionActions,
+        loader?: boolean
+    ) => (...args: any[]) => Promise<void>;
+    tooltipTitle?: string;
+}
+
+export function ShareOption({
```
abhinavkgrd, line 15: Following the convention of naming the component fully ( `ShareQuickOption` ) removes the confusion when importing it.

Else there can be cases where we can have `ShareCollectionOption` and `ShareQuickOption` both named as `ShareOption` and then will have to look into where the file is coming from to help figure out which one to import.

I know not a big issue... But just simply naming them completely fixes that import issue from ever happening. Also, `vscode` fix import suggestion will also never mistakenly import the wrong thing

src/components/Collections/CollectionOptions/QuickOptions/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+                <ShareOption
+                    handleCollectionAction={handleCollectionAction}
+                    tooltipTitle={
+                        /*: collectionSummaryType ===
+                    CollectionSummaryType.incomingShare
+                  ? constants.SHARING_DETAILS*/
+                        collectionSummaryType ===
+                            CollectionSummaryType.outgoingShare ||
+                        collectionSummaryType ===
+                            CollectionSummaryType.sharedOnlyViaLink
```
abhinavkgrd, line 57: same here we can move the logic figure out the `tooltipTitle` to the component

src/utils/collection/index.ts
```diff
@@ -195,7 +195,20 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
 export const shouldShowOptions = (type: CollectionSummaryType) => {
     return !OPTIONS_NOT_HAVING_COLLECTION_TYPES.has(type);
 };
-
+export const showTrashQuickOption = (type: CollectionSummaryType) => {
+    return type === CollectionSummaryType.trash;
+};
+export const showDownloadQuickOption = (type: CollectionSummaryType) => {
+    return !(type === CollectionSummaryType.trash);
```
abhinavkgrd, line 202: `type !== CollectionSummaryType.trash ` 
simpler?

src/utils/collection/index.ts
```diff
@@ -195,7 +195,20 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
...
+};
+export const showShareQuickOption = (type: CollectionSummaryType) => {
+    return !(
+        type === CollectionSummaryType.trash ||
```
abhinavkgrd, line 206: can say when to show it... The current logic will start showing it for any new type, we add by default. 

if we decide where to show it then... rather than where to not... I will never to mistakenly shown for cases we didn't intend to

src/utils/collection/index.ts
abhinavkgrd: also as suggested in the last comment... let's say which to show for... rather than for which to not show

src/components/Collections/CollectionOptions/QuickOptions/ShareOption.tsx
abhinavkgrd: Ah, I think, you misunderstood this. I wanted the component to be named `ShareQuickOption`, but it's still called `shareOption` and you have renamed the filename to `ShareQuickOption`. Lets fix that

src/components/Collections/CollectionOptions/QuickOptions/index.tsx
abhinavkgrd: same, wanted the component to be renamed not the file

src/components/Collections/CollectionOptions/QuickOptions/MoveToTrashOption.tsx
abhinavkgrd: this is not done yet no?, the component is still called `moveToTrashOption` ?

src/utils/collection/index.ts
```diff
@@ -194,7 +194,30 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
...
 };
-
+export const showTrashQuickOption = (type: CollectionSummaryType) => {
+    return type === CollectionSummaryType.trash;
+};
+export const showDownloadQuickOption = (type: CollectionSummaryType) => {
+    return (
+        type === CollectionSummaryType.folder ||
+        type === CollectionSummaryType.album ||
+        type === CollectionSummaryType.all ||
```
abhinavkgrd, line 204: I don't think we, showed any option for `all` section.

src/utils/collection/index.ts
```diff
@@ -194,7 +194,30 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
...
+        type === CollectionSummaryType.outgoingShare ||
+        type === CollectionSummaryType.sharedOnlyViaLink ||
+        type === CollectionSummaryType.archived
+    );
+};
+export const showShareQuickOption = (type: CollectionSummaryType) => {
+    return (
+        type === CollectionSummaryType.folder ||
+        type === CollectionSummaryType.album ||
+        type === CollectionSummaryType.all ||
```
abhinavkgrd, line 215: same here

src/utils/collection/index.ts
```diff
@@ -194,7 +194,30 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
...
+        type === CollectionSummaryType.outgoingShare ||
+        type === CollectionSummaryType.sharedOnlyViaLink ||
+        type === CollectionSummaryType.archived
```
abhinavkgrd, line 218: same here

src/utils/collection/index.ts
```diff
@@ -194,7 +194,30 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
 export const shouldShowOptions = (type: CollectionSummaryType) => {
     return !OPTIONS_NOT_HAVING_COLLECTION_TYPES.has(type);
 };
-
+export const showTrashQuickOption = (type: CollectionSummaryType) => {
```
abhinavkgrd, line 197: also lets rename this to `showEmptyTrashQuickOption`

src/utils/collection/index.ts
```diff
@@ -194,7 +194,30 @@ export const isSystemCollection = (type: CollectionSummaryType) => {
...
+};
+export const showDownloadQuickOption = (type: CollectionSummaryType) => {
+    return (
+        type === CollectionSummaryType.folder ||
+        type === CollectionSummaryType.album ||
+        type === CollectionSummaryType.all ||
+        type === CollectionSummaryType.incomingShare ||
+        type === CollectionSummaryType.outgoingShare ||
+        type === CollectionSummaryType.sharedOnlyViaLink ||
+        type === CollectionSummaryType.archived
```
abhinavkgrd, line 208: `archive` also doesn't have the option

abhinavkgrd (approved):

PR #941, opened 2023-02-10
Remove deprecated search API

mnvr:
## Test Plan

Can build and run locally. Also tested search. Didn't specifically test this code path but that should be fine because this function currently isn't even called because its caller (`getLocationSuggestions`) is unused,

### DIFF
--- a/src/services/searchService.ts
+++ b/src/services/searchService.ts
@@ -22,7 +22,6 @@ import { Person, Thing } from 'types/machineLearning';
 import { getUniqueFiles } from 'utils/file';
 import { User } from 'types/user';
 import { getData, LS_KEYS } from 'utils/storage/localStorage';
-import { addLogLine } from 'utils/logging';
 
 const DIGITS = new Set(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']);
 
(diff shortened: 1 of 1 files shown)


src/services/searchService.ts
```diff
@@ -205,23 +201,10 @@ function parseHumanDate(humanDate: string): DateValue[] {
...
-            },
-            {
-                'X-Auth-Token': getToken(),
-            }
-        );
-        return resp.data.results ?? [];
-    } catch (e) {
-        logError(e, 'location search failed');
-    }
+    addLogLine(`attempting to use unimplemented search API`);
```
abhinavkgrd, line 207: we should log to sentry also, just in case

mnvr: I wasn't sure if we should use

1.    ` logError(Error('attempting to use unimplemented search API'));`

or 

2.    ` logError(Error(), 'attempting to use unimplemented search API');

I found one example in code that did #2, so I did the same.

abhinavkgrd: ya two works

abhinavkgrd (dismissed):

mnvr:
Resolved merge conflicts, and re-ran local build. Please reapprove 🙏

abhinavkgrd (approved):

PR #966, opened 2023-03-02
Add comments

vishnukvmd:


src/utils/crypto/index.ts
```diff
@@ -58,6 +58,10 @@ export async function generateKeyAttributes(
     return { keyAttributes, masterKey };
 }
 
+// We encrypt existingKeyAttributes with a key derived from the passphrase (with
+// Interactive mem and ops limits) to avoid saving them to local storage in
```
abhinavkgrd, line 62: > We encrypt existingKeyAttributes

actually we create complete new keyAttributes 

This was done, more because the decryption originalKeyAttributes to get the masterPassword was slow, so we created an IntermeditateKeyAttributes (for local use whenever the user reopens the web app), which is masterKey encrypted with less strict mem and ops limits.

abhinavkgrd (approved):

PR #983, opened 2023-03-19
Display ente authenticator codes

ua741:
## Description
**Pending**

- [x] Fetch and render codes stored on server.
- [x] Verify current and next TOTP codes

---- 

- Planning to just add code to render the ente Auth codes with no support to edit/add/delete codes.
- Whether we will reuse flutter code or go via react/electronJS route for ente Auth is still TBD.
- In case of react/electron, whether we will ship a different package or render ente auth codes within ente Photos app is TDB.


src/services/authenticator/authenticatorService.ts
```diff
@@ -0,0 +1,102 @@
...
+                .map(async (entity) => {
+                    const decryptedCode = await cryptoWorker.decryptMetadata(
+                        entity.encryptedData,
+                        entity.header,
+                        authentitorKey
+                    );
+                    try {
+                        return Code.fromRawData(entity.id, decryptedCode);
+                    } catch (e) {
+                        console.log(
```
vishnukvmd, line 34: Note: have to remove this before merging

src/pages/authenticator/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    })
+                    .catch((err) => {
+                        if (err.message === CustomError.KEY_MISSING) {
+                            router.push({
+                                pathname: PAGES.CREDENTIALS,
+                                query: { redirectPage: PAGES.AUTHENICATOR },
+                            });
+                            return;
+                        }
+                        console.error('something wrong here', err);
```
vishnukvmd, line 29: On mobile, we decided to not log the errors for auth, since the underlying error could bubble up sensitive data

src/pages/authenticator/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    });
+            } catch (error) {
+                console.error('something wrong where asdas');
```
vishnukvmd, line 32: 👀

src/pages/authenticator/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    return (
+        <div
+            style={{
+                display: 'flex',
+                flexDirection: 'column',
+                alignItems: 'center',
+                justifyContent: 'flex-start',
+            }}>
+            <div style={{ marginBottom: '2rem' }} />
+            <h2>ente Authenticator</h2>
```
vishnukvmd, line 82: `ente` is already in the title bar, let's just keep `Authenticator`

src/services/authenticator/authenticatorService.ts
```diff
@@ -0,0 +1,97 @@
...
+                .map(async (entity) => {
+                    const decryptedCode = await cryptoWorker.decryptMetadata(
+                        entity.encryptedData,
+                        entity.header,
+                        authentitorKey
+                    );
+                    try {
+                        return Code.fromRawData(entity.id, decryptedCode);
+                    } catch (e) {
+                        console.log('failed to parse code', e, entity.id);
```
vishnukvmd, line 34: Let's not log `e`?

src/constants/pages/index.ts
```diff
@@ -14,4 +14,5 @@ export enum PAGES {
     SHARED_ALBUMS = '/shared-albums',
     // ML_DEBUG = '/ml-debug',
     DEDUPLICATE = '/deduplicate',
+    AUTHENICATOR = '/authenticator',
```
vishnukvmd, line 17: Let's switch to `/auth`?

abhinavkgrd (approved):

src/pages/authenticator/index.tsx
ua741: In the end, we ended up adding logs to debug and fix customer issues :D.
Can remove it if we want. But might need to back it back again in the future just in case of things need to be debugged.

src/services/authenticator/authenticatorService.ts
ua741: Ack.
Btw, I don't think exception will contain any sensitive data from my limited testing of various cases. 
From my experience, most libs  throw error without the input params in the msg. Now, if we think about it, it's a good practice as that can result in accidental leaks during error logging.

src/components/Authenicator/OTPDisplay.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    counter: 0,
+                    algorithm: codeInfo.algorithm,
+                });
+                setCode(hotp.generate());
+                setNextCode(hotp.generate({ counter: 1 }));
+            }
+        } catch (err) {
+            setCodeErr(err.message);
+        }
+    };
```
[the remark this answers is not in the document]
ua741: Didn't understood this part. I want to accomplish following behaviour:

1. Immediately call `generateCode` on the rendering of component
2. Call `generateCode` every 1 seconds so that after pre-defined period (30/60s etc), new code is generated.

Given that, can you suggest what changes shall I make?

src/types/authenticator/code.ts
```diff
@@ -0,0 +1,182 @@
...
+            .replace(/:/g, '%3A')
+            .replaceAll('\r', '');
+        if (santizedRawData.startsWith('"')) {
+            santizedRawData = santizedRawData.substring(1);
+        }
+        if (santizedRawData.endsWith('"')) {
+            santizedRawData = santizedRawData.substring(
+                0,
+                santizedRawData.length - 1
+            );
```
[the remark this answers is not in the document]
ua741: Shall we switch to this after I come back. Right now, have tested all functionality with current package. Switching it might break it and will require remote debugging with Vishnu

src/components/Authenicator/OTPDisplay.tsx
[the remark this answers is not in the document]
abhinavkgrd: What I meant was that, 

we can have one function, `createCodeGeneration` which would just create and return the generator instance 
```
const createCodeGeneration = () => {
        if (codeInfo.type.toLowerCase() === 'totp') {
            return new TOTP({
                secret: codeInfo.secret,
                algorithm: codeInfo.algorithm ?? Code.defaultAlgo,
                period: codeInfo.period ?? Code.defaultPeriod,
                digits: codeInfo.digits ?? Code.defaultDigits,
            });
        } else if (codeInfo.type.toLowerCase() === 'hotp') {
            return new HOTP({
                secret: codeInfo.secret,
                counter: 0,
                algorithm: codeInfo.algorithm,
            });
        }
    };
```

and with that we can generate the currentCode by calling `generator.generate()` and for the next code we could have a helper function like `generateNextCodes`

```
 const generateNextCodes = (generator: TOTP | HOTP, type: Code['type']) => {
        if (type.toLowerCase() === 'totp') {
            const currentTime = new Date().getTime();
            return generator.generate({
                timestamp: currentTime + codeInfo.period * 1000,
            });
        } else if (codeInfo.type.toLowerCase() === 'hotp') {
            return generator.generate({ counter: 1 });
        }
    };
```

now the useEffect could be updated to and the two function `setupGenerator` and `generateNextCode` could be moved to util/service file 

```

const updateCodes = (generator: TOTP | HOTP) => {
        try {
            const code = generator.generate();
            const nextCode = generateNextCodes(generator, codeInfo.type);
            setCode(code);
            setNextCode(nextCode);
        } catch (err) {
            setCodeErr(err.message);
        }
    };

    useEffect(() => {
        let interval;
        try {
            const generator = createCodeGeneration();
            updateCodes(generator);
            interval = setInterval(() => {
                updateCodes(generator);
            }, generateCodeInterval);
        } catch (err) {
            setCodeErr(err.message);
        }
        return () => {
            clearInterval(interval);
        };
    }, []);

```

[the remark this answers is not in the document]
abhinavkgrd: This can be improved further, as we are unneccesarily trying to udpate Code every second. 

Which we can fix by updating the code something like this.

```

const setupCodeUpdateLoop = (generator: TOTP | HOTP) => {
        const microSecondsInPeriod = codeInfo.period * 1000;
        const timeToNextCode =
            microSecondsInPeriod -
            ((new Date().getTime() * 1000) % microSecondsInPeriod);

        let interval: NodeJS.Timeout;
        const timeout = setTimeout(() => {
            updateCodes(generator);
            interval = setInterval(() => {
                updateCodes(generator);
            }, microSecondsInPeriod);
        }, timeToNextCode);
        return {
            cleanUp: () => {
                clearTimeout(timeout);
                clearInterval(interval);
            },
        };
    };

    useEffect(() => {
        try {
            const generator = createCodeGenerator();
            updateCodes(generator);
            const { cleanUp } = setupCodeUpdateLoop(generator);
            return () => {
                cleanUp();
            };
        } catch (err) {
            setCodeErr(err.message);
        }
    }, []);
```

src/pages/authenticator/index.tsx
abhinavkgrd, line 1: we will also need to rename this file to `auth/index.tsx` (as we have updated the auth route)

abhinavkgrd:
https://user-images.githubusercontent.com/46242073/227700646-81503535-8d1b-4df3-b490-16a470559f3b.mov

Found a small issue. The code doesn't update in sync with the TimerProgress.



This is because both working on different intervals.

The fix could be to setup the timers such that they update at the exact time the code.

Currently, it is within the `1 sec` window of it. But not actually in sync.

we can use something like this to fix it.

```
 useEffect(() => {
        // this is to set the initial code and nextCode on component mount
        generateCodes();
        const codeType = codeInfo.type;
        const codePeriodInMs = codeInfo.period * 1000;
        const timeToNextCode =
            codePeriodInMs - (new Date().getTime() % codePeriodInMs);
        const intervalId = null;
        // wait until we are at the start of the next code period,
        // and then start the interval loop
        setTimeout(() => {
            // we need to call generateCodes() once before the interval loop
            // to set the initial code and nextCode
            generateCodes();
            codeType.toLowerCase() === 'totp' ||
            codeType.toLowerCase() === 'hotp'
                ? setInterval(() => {
                      generateCodes();
                  }, codePeriodInMs)
                : null;
        }, timeToNextCode);

        return () => {
            if (intervalId) clearInterval(intervalId);
        };
    }, [codeInfo]);
```

src/components/Authenicator/OTPDisplay.tsx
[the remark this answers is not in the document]
ua741: (discussed offline, no action item needed)

abhinavkgrd (approved):

PR #988, opened 2023-03-22
Simplify export Flows

abhinavkgrd:
## Description

- removed export size and added total file count info
- updated layout
-  removed separately retry failed export button, same resync button will trigger export for all unexported files( including those failed during last export )
- removed pausing/resume logic, export now always resumes from the last export state automatically
- deprecated `queuedFiles` and `progress`  properties from `exportRecord`

## Test Plan

pending

### DIFF
--- a/src/utils/file/index.ts
+++ b/src/utils/file/index.ts
@@ -578,3 +578,11 @@ export function getLatestVersionFiles(files: EnteFile[]) {
         (file) => !file.isDeleted
     );
 }
+
+export function getUserPersonalFiles(files: EnteFile[]) {
+    const user: User = getData(LS_KEYS.USER);
+    if (!user?.id) {
+        throw Error('user missing');
+    }
+    return files.filter((file) => file.ownerID === user.id);
+}
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -300,7 +300,7 @@
     "EXPORT_DATA": "Export data",
     "SELECT_FOLDER": "Select folder",
     "DESTINATION": "Destination",
-    "EXPORT_FILE_COUNT": "File count",
+    "FILE_COUNT": "File count",
     "START": "Start",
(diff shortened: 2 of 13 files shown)


src/utils/file/index.ts
```diff
@@ -578,3 +579,20 @@ export function getLatestVersionFiles(files: EnteFile[]) {
         (file) => !file.isDeleted
     );
 }
+
+export function getUserPersonalFiles(files: EnteFile[]) {
```
ua741, line 583: can this logic be part of `getTotalFileCount`? I don't see us reusing it anywhere else as of now. 

```ts
export const getUserPersonalFileCount = async (): Promise<number> => {
    try {
        const user: User = getData(LS_KEYS.USER);
        if (!user?.id) {
            throw Error('user missing');
        }
        const files: EnteFile[] = await getLocalFiles();
        const userPersonalFiles = files.filter((file) => file.ownerID === user.id);
        return userPersonalFiles.length;
    } catch (e) {
        logError(e, 'getUserPersonalFileCount failed');
        throw e;
    }
};
```

src/utils/file/index.ts
```diff
@@ -578,3 +579,20 @@ export function getLatestVersionFiles(files: EnteFile[]) {
...
+    const user: User = getData(LS_KEYS.USER);
+    if (!user?.id) {
+        throw Error('user missing');
+    }
+    return files.filter((file) => file.ownerID === user.id);
+}
+
+export const getTotalFileCount = async () => {
```
ua741, line 591: Also, will this method return count of collection_files or files? I am not sure what do we need it return :)

abhinavkgrd: It will return the count of `collection_files`

abhinavkgrd: as, Files are not deduped during export, so showing `collection_files` count made more sense

ua741 (approved):

PR #1008, opened 2023-04-01
Remove unsed strings

abhinavkgrd:
## Description

- added script to purge unused string from translation json  

## Test Plan
tested by checking all the removed strings are actually not used in the code


purge_unused_strings.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
vishnukvmd, line 1: Let's place this inside a `scripts` folder?

vishnukvmd (dismissed):

vishnukvmd (approved):

PR #1045, opened 2023-04-13
revert offset handling

abhinavkgrd:
## Description

Found images where the camera has stored the date in local time without providing an offset, so safer to just assume that the time is local time. This will be correct for most people unless travelling across timezones

## Test Plan

tested locally and added the file to the sample dataset


ua741 (approved):

ua741:
Can we update our test data set for this change?

abhinavkgrd:
> Can we update our test data set for this change?

Yepp already did

PR #1067, opened 2023-04-18
Monorepo

httpjamesm:
## Description

This pull request introduces changes to convert the `photos-web` project into a Turborepo monorepo. The conversion aims to improve the development experience and optimize build times using Turborepo's caching and parallelization features.

Key changes include:

1. Setting up Turborepo: The project now uses Turborepo for managing tasks and dependencies across workspaces.

2. Monorepo structure: The project has been restructured into a monorepo, with separate workspaces for different parts of the application.

3. Configuration updates: The `turbo.json` and `package.json` files have been updated to accommodate the new monorepo structure and to utilize Turborepo's features.

4. TypeScript configuration: The `tsconfig.json` files have been adjusted to work with the new monorepo setup.

## Test Plan

1. Move your `.env` to `apps/photos`.
2. Run `yarn dev` at the root of the monorepo.
3. Confirm the app runs properly.


apps/photos/next.config.js
```diff
@@ -48,6 +48,8 @@ module.exports = (phase) =>
                 '@mui/material',
                 '@mui/system',
                 '@mui/icons-material',
+                'tsconfig',
+                'eslint-config-custom'
```
abhinavkgrd, line 52: we can name this `eslint-config-ente`?

apps/photos/public/netlify.toml
abhinavkgrd, line 1: we can delete this...not needed anymore

apps/photos/thirdparty/ffmpeg-wasm/.eslintrc
abhinavkgrd, line 1: Seems like the submodule setup broke and ffmpeg-wasm's code is being consider as our repo code ?

I think we will need to update the module path in `.submodules`.

same for `photoswipe` too

package.json
```diff
@@ -1,119 +1,123 @@
...
+        "lint": "turbo run lint",
+        "albums": "next dev -p 3002",
+        "auth": "next dev -p 3004",
+        "export": "next export",
+        "build-analyze": "ANALYZE=true next build",
+        "start": "next start",
+        "prepare": "husky install"
+    },
+    "dependencies": {
+        "@date-io/date-fns": "^2.14.0",
```
abhinavkgrd, line 17: we can remove these dependencies from here now?

packages/eslint-config-custom/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    extends: ["next", "turbo", "prettier"],
+    rules: {
+      "@next/next/no-html-link-for-pages": "off",
```
abhinavkgrd, line 4: 😬 , we are not using this config...seems like an example config from the turbo repo?

abhinavkgrd: also, it doesn't seem like we have setup eslint in the photos app? I can just see the package in package.json but no usage of it.

abhinavkgrd: our eslint config https://github.com/httpjamesm/photos-web/blob/bc5c1c7f162da1dd0d970dde0d7160f030d754c8/.eslintrc.json

packages/tsconfig/react-library.json
abhinavkgrd, line 1: We don't need this no?

packages/tsconfig/nextjs.json
abhinavkgrd, line 1: All our apps are gonna be nextJS only. So, I don't think a layer over base config would be required

we can I think just use the base config

package.json
```diff
@@ -1,119 +1,27 @@
...
+    "scripts": {
+        "build": "turbo run build && turbo run export",
+        "dev": "turbo run dev --parallel",
+        "lint": "turbo run lint",
+        "albums": "next dev -p 3002",
+        "auth": "next dev -p 3004",
+        "export": "next export",
+        "build-analyze": "ANALYZE=true next build",
+        "start": "next start",
+        "prepare": "husky install"
```
abhinavkgrd, line 14: I assume it would need to have `husky` installed to be at the root level. for this to work?

Also we currently lint the app on `pre-commit ` hook. Can you check how that will work, such that the app-specific eslint-config is used

packages/eslint-config-ente/index.js
```diff
@@ -0,0 +1,63 @@
+module.exports = {
+    extends: [
+        'next',
+        'turbo',
+        'prettier',
+        'plugin:@typescript-eslint/recommended',
+        'plugin:@typescript-eslint/recommended-requiring-type-checking',
```
abhinavkgrd, line 7: we had a different set of imports earlier. lets use that to stay consistent and not change the behaviour. 

also `prettier`  needs to be at the end for it to work properly.

packages/eslint-config-ente/index.js
```diff
@@ -0,0 +1,63 @@
...
+    ],
```
abhinavkgrd, line 8: was 

`"plugins": ["@typescript-eslint"],`

removed intentionally?

package.json
httpjamesm: I've fixed linting and made sure they use the correct configs, so it should work. The husky pre-commit command simply runs the lint-staged command, which I assume uses the same logic as the normal lint command.

.gitmodules
```diff
@@ -1,8 +1,8 @@
...
-	path = thirdparty/ffmpeg-wasm
+[submodule "apps/photos/thirdparty/ffmpeg-wasm"]
+	path = apps/photos/thirdparty/ffmpeg-wasm
 	url = https://github.com/abhinavkgrd/ffmpeg.wasm.git
-	branch = single-thread
+    branch = master 
+[submodule "apps/photos/thirdparty/photoswipe"]
+	path = apps/photos/thirdparty/photoswipe
+	url = https://github.com/ente-io/PhotoSwipe.git
+	branch = single-thread
```
abhinavkgrd, line 8: new line missing 😅

apps/photos/next.config.js
```diff
@@ -48,6 +48,8 @@ module.exports = (phase) =>
                 '@mui/material',
                 '@mui/system',
                 '@mui/icons-material',
+                'tsconfig',
```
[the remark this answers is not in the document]
abhinavkgrd: we need to rename this no ?

apps/photos/package.json
```diff
@@ -0,0 +1,121 @@
...
+        "lint-staged": "^11.1.2",
+        "prettier": "2.3.2",
+        "typescript": "^4.1.3",
+        "@ente/tsconfig": "*",
+        "@ente/eslint-config": "*"
+    },
+    "standard": {
+        "parser": "babel-eslint"
+    }
+}
```
abhinavkgrd, line 121: prettier is not working  👀 ? there are lots of new lines missing.

apps/photos/thirdparty/ffmpeg-wasm
abhinavkgrd, line 1: I think something got changed in the submodule...there should be no diff for it.

package.json
abhinavkgrd: but shouldn't it be in the `package.json` devDependency?

Also, how is lint-staged running that is also not added as a dependency?

also the `.lintstagedrc.js`  runs `next lint` and `yarn prettier` commands how are they present without them being added as dependency?

packages/eslint-config-ente/index.js
abhinavkgrd: this is not resolved

PR #1098, opened 2023-05-08
Handle zip drop on desktop

abhinavkgrd:
## Description

intercept drag and drop and check if the file is zip then try reading zip content.

## Test Plan

- tested by uploading test zip dataset
- tested normal upload

### DIFF
--- a/src/components/Upload/Uploader.tsx
+++ b/src/components/Upload/Uploader.tsx
@@ -219,7 +219,7 @@ export default function Uploader(props: Props) {
                     try {
                         addLogLine(`uploading dropped files from desktop app`);
                         // check and parse dropped files which are zip files
-                        const electronFiles = [] as ElectronFile[];
+                        let electronFiles = [] as ElectronFile[];
                         for (const file of props.dragAndDropFiles) {
                             if (file.name.endsWith('.zip')) {
                                 const zipFiles =
@@ -229,7 +229,7 @@ export default function Uploader(props: Props) {
                                 addLogLine(
                                     `zip file - ${file.name} contains ${zipFiles.length} files`
                                 );
-                                electronFiles.push(...zipFiles);
+                                electronFiles = [...electronFiles, ...zipFiles];
                             } else {
                                 // type cast to ElectronFile as the file is dropped from desktop app
                                 // type file and ElectronFile should be interchangeable, but currently they have some differences.
--- a/src/components/Upload/Uploader.tsx
+++ b/src/components/Upload/Uploader.tsx
@@ -219,11 +219,9 @@ export default function Uploader(props: Props) {
                     try {
                         addLogLine(`uploading dropped files from desktop app`);
                         // check and parse dropped files which are zip files
-
                         const electronFiles = [
                             ...(await Promise.all(
                                 props.dragAndDropFiles.map(async (file) => {
-                                    addLogLine(`dropped file - ${file.name}`);
                                     if (file.name.endsWith('.zip')) {
                                         const zipFiles =
                                             await importService.getElectronFilesFromGoogleZip(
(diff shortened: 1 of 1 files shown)


ua741:
What will be the behavior if user drops a folder which has mix of photos and zip files?

abhinavkgrd:
> What will be the behavior if user drops a folder which has mix of photos and zip files?

the zip file will be parsed and its files will be added to the list along with normal files

so, if the user drops `[sampleZip.zip, sampleFile.png]`

then we would convert it to  `[ sampleZipFile1.png, sampleZipFile2.mp4, sampleFile.png]`

src/components/Upload/Uploader.tsx
```diff
@@ -214,8 +214,45 @@ export default function Uploader(props: Props) {
...
+            if (isElectron()) {
+                const main = async () => {
+                    try {
+                        addLogLine(`uploading dropped files from desktop app`);
+                        // check and parse dropped files which are zip files
+
+                        const electronFiles = [
+                            ...(await Promise.all(
+                                props.dragAndDropFiles.map(async (file) => {
+                                    addLogLine(`dropped file - ${file.name}`);
```
ua741, line 226: Do we really need to log this information here?

abhinavkgrd: No, we can drop it. will remove it

src/components/Upload/Uploader.tsx
```diff
@@ -214,8 +214,45 @@ export default function Uploader(props: Props) {
...
-            setWebFiles(props.dragAndDropFiles);
+            if (isElectron()) {
+                const main = async () => {
+                    try {
+                        addLogLine(`uploading dropped files from desktop app`);
+                        // check and parse dropped files which are zip files
+
+                        const electronFiles = [
+                            ...(await Promise.all(
+                                props.dragAndDropFiles.map(async (file) => {
```
ua741, line 224: Just a side note: all of these .map, prormise.all functions will slowly slowly start adding up when it comes to performance.
Here, i would return a for loop to identify zip files, and just zipped them to add via promise.

abhinavkgrd: But then also we would need to have an async for loop, which would wait for the zipping to complete no?

The same is happening currently too no? We await if the file is a zip to unzip, otherwise, we just push return the file, in which case it just acts like a sync function.

abhinavkgrd: The for-loop version would be a cleaner,

src/components/Upload/Uploader.tsx
```diff
@@ -214,8 +214,43 @@ export default function Uploader(props: Props) {
...
+                                        const zipFiles =
+                                            await importService.getElectronFilesFromGoogleZip(
+                                                (file as any).path
+                                            );
+                                        addLogLine(
+                                            `zip file - ${file.name} contains ${zipFiles.length} files`
+                                        );
+                                        return zipFiles;
+                                    } else {
+                                        return file as unknown as ElectronFile;
```
ua741: What's the difference between.. Can we add a comment why we have to cast it as `unknown` first?
`return file  as ElectronFile;` vs `return file as unknown as ElectronFile`;

ua741 (dismissed):

src/components/Upload/Uploader.tsx
abhinavkgrd: We earlier created a separate type for files coming from Electron(desktop), but mostly they can be used interchangeably.

I didn't want to do the complete type and make this PR too big and bloated with non-related changes.

So, for now, I did `file as unknown as ElectronFile;` which just is a way to tell the compiler to ignore the type difference between the two types and type cast file as ElectronFile.


Yes adding a comment

ua741 (approved):

PR #1106, opened 2023-05-10
Hidden support

abhinavkgrd:
## Description
- Added Sidebar Option to open Hidden Section after password authentication
- Added hidden section with option same as uncategorized section
- Added support to sync hidden files 
- Added Support to Hide and unhide files.
- Hide option available in All section + collection view 
- Handle downloading multiple default hidden collection

## Test Plan

- Verified Hidden section Sidebar authentication working properly
- Tested delete account flow authentication working properly 
- Tested syncing hidden files from remote
- Tested hiding and unhiding files
- Confirmed hidden files not visible during search/dedupe


apps/photos/src/utils/collection/index.ts
```diff
@@ -260,12 +272,27 @@ export function isValidMoveTarget(
...
+export function isValidReplacementAlbum(
+    collection: Collection,
+    user: User,
+    wantedCollectionName: string
+) {
+    return (
+        collection.name === wantedCollectionName &&
+        (collection.type === CollectionType.album ||
+            collection.type === CollectionType.folder) &&
+        !isHiddenCollection(collection) &&
```
ua741, line 290: Why add this check?
Suppose if I have a hidden folder on my machine. I drag it and mark it the folder has hidden. Next time when I am dragging the same folder, we shouldn't be creating new album. Adding a check against `defaultHiddenFolder` and `isIncomingShare` should be good enough.

apps/photos/src/services/collectionService.ts
```diff
@@ -173,24 +179,28 @@ const getCollections = async (
...
-    return getNonHiddenCollections(collections);
+    return includeHidden ? collections : getNonHiddenCollections(collections);
 };
 
 export const getCollectionUpdationTime = async (): Promise<number> =>
     (await localForage.getItem<number>(COLLECTION_UPDATION_TIME)) ?? 0;
 
-export const syncCollections = async () => {
-    const localCollections = await getLocalCollections();
+export const syncCollections = async (includeHidden = false) => {
```
ua741, line 193: What are the cases when we wouldn't want to sync `hiddenCollections`?

apps/photos/src/utils/file/index.ts
```diff
@@ -501,12 +486,12 @@ export const createTypedObjectURL = async (blob: Blob, fileName: string) => {
     return URL.createObjectURL(new Blob([blob], { type: type.mimeType }));
 };
 
-export const getUserOwnedNonTrashedFiles = (files: EnteFile[]) => {
+export const getUserOwnedFiles = (files: EnteFile[]) => {
     const user: User = getData(LS_KEYS.USER);
     if (!user?.id) {
         throw Error('user missing');
     }
-    return files.filter((file) => file.isTrashed || file.ownerID === user.id);
```
ua741, line 509: Can you tell me why was this change done.. How are we now handling trashedFiles?

apps/photos/src/utils/collection/index.ts
abhinavkgrd: yes yes, it does that only, we currently only have default hidden collection and `isHiddenCollection` checks for it is the defaultHiddenCollection. Will update the function name

ua741 (dismissed):

ua741 (approved):

PR #1138, opened 2023-05-27
Fix grouping based on hashes on dedupe page 

abhinavkgrd:
## Description

^ title 

## Test Plan

tested locally by uploading a file multiple copies, some with hash and some without hash


apps/photos/src/components/PhotoList.tsx
```diff
@@ -353,7 +356,13 @@ export function PhotoList({
                     ) !== currentFileSize ||
                     (deduplicateContext.clubSameTimeFilesOnly &&
                         displayFiles[lastFileIndex].metadata.creationTime !==
-                            currentCreationTime)
+                            currentCreationTime) ||
```
vishnukvmd, line 359: Refactor this if condition?

vishnukvmd (approved):

PR #1145, opened 2023-05-29
[WIP] Merge master in locker

ua741:
## Description
Issues to be fixex:
- Eslint seems to be missing on locker, need to fix those issues.
- Also locker app is directly importing from `photos` app, that's also causing compiler to cry


abhinavkgrd:
james has merged `main` into `locker` and it is working properly

closing this

PR #1151, opened 2023-05-30
Added download album option on shared-albums

abhinavkgrd:
## Description

image

## Test Plan

tested locally


ua741:
If the download is disabled in public collection, will this button be hidden?

abhinavkgrd:
> If the download is disabled in public collection, will this button be hidden?

ahh, thanks!!. nice catch...fixed it

ua741 (approved):

PR #1159, opened 2023-06-05
Email-suggestion-collection-sharing

jubitjohn:
## Description

Added email suggestion list  to album sharing using Autocomplete, which will provide the users with a list of email with who they have previosly shared of the ones shared with them

### DIFF
--- a/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
+++ b/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
@@ -20,9 +20,9 @@ export default function EmailShare({ collection }) {
     useEffect(() => {
         const getUpdatedOptionsList = async () => {
             const ownerUser: User = getData(LS_KEYS.USER);
-            const collection_list = getLocalCollections();
-            const family_list = getLocalFamilyData();
-            const result = await collection_list;
+            const collectionList = getLocalCollections();
+            const familyList = getLocalFamilyData();
+            const result = await collectionList;
             const emails = result.flatMap((item) => {
                 const shareeEmails = item.sharees.map((sharee) => sharee.email);
                 if (item.owner.email) {
@@ -33,10 +33,8 @@ export default function EmailShare({ collection }) {
             });
 
             // adding family members
-            if (family_list) {
-                const family = family_list.members.map(
-                    (member) => member.email
-                );
+            if (familyList) {
+                const family = familyList.members.map((member) => member.email);
                 emails.push(...family);
             }
 
--- a/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
+++ b/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
@@ -17,9 +17,9 @@ export default function EmailShare({ collection }) {
     const [updatedOptionsList, setUpdatedOptionsList] = useState(['']);
     let updatedList = [];
     useEffect(() => {
-        const owner_user1: User = getData(LS_KEYS.USER);
-        const collection_list = getLocalCollections();
         const getUpdatedOptionsList = async () => {
+            const ownerUser: User = getData(LS_KEYS.USER);
+            const collection_list = getLocalCollections();
             const result = await collection_list;
             const emails = result.flatMap((item) => {
                 const shareeEmails = item.sharees.map((sharee) => sharee.email);
@@ -36,7 +36,7 @@ export default function EmailShare({ collection }) {
                 (email) =>
                     !collection.sharees
                         .map((sharees) => sharees.email)
-                        .includes(email) && email !== owner_user1.email
+                        .includes(email) && email !== ownerUser.email
             );
 
             setUpdatedOptionsList(filteredList);
--- a/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
+++ b/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
@@ -10,6 +10,7 @@ import { handleSharingErrors } from 'utils/error/ui';
 import { getData, LS_KEYS } from 'utils/storage/localStorage';
 import { CollectionShareSharees } from './sharees';
 import { getLocalCollections } from 'services/collectionService';
+import { getLocalFamilyData } from 'utils/user/family';
 
 export default function EmailShare({ collection }) {
     const galleryContext = useContext(GalleryContext);
@@ -20,6 +21,7 @@ export default function EmailShare({ collection }) {
         const getUpdatedOptionsList = async () => {
             const ownerUser: User = getData(LS_KEYS.USER);
             const collection_list = getLocalCollections();
+            const family_list = getLocalFamilyData();
             const result = await collection_list;
             const emails = result.flatMap((item) => {
                 const shareeEmails = item.sharees.map((sharee) => sharee.email);
@@ -29,6 +31,14 @@ export default function EmailShare({ collection }) {
                     return shareeEmails;
                 }
             });
+
+            // adding family members
+            if (family_list) {
+                const family = family_list.members.map(
+                    (member) => member.email
+                );
+                emails.push(...family);
+            }
 
             updatedList = Array.from(new Set(emails));
 
--- a/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
+++ b/apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
@@ -2,16 +2,56 @@ import SingleInputForm, {
     SingleInputFormProps,
 } from 'components/SingleInputForm';
 import { GalleryContext } from 'pages/gallery';
-import React, { useContext } from 'react';
+import React, { useContext, useState, useEffect } from 'react';
 import { t } from 'i18next';
 import { shareCollection } from 'services/collectionService';
 import { User } from 'types/user';
 import { handleSharingErrors } from 'utils/error/ui';
 import { getData, LS_KEYS } from 'utils/storage/localStorage';
 import { CollectionShareSharees } from './sharees';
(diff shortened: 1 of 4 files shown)


apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -88,40 +90,85 @@ export default function SingleInputForm(props: SingleInputFormProps) {
...
+                            {...(props.hiddenLabel
+                                ? { placeholder: props.placeholder }
+                                : { label: props.placeholder })}
+                            value={values.inputValue}
+                            onChange={handleChange('inputValue')}
+                            error={Boolean(errors.inputValue)}
+                            helperText={errors.inputValue}
+                            disabled={loading}
+                            autoFocus={!props.disableAutoFocus}
+                            autoComplete={props.autoComplete}
```
ua741, line 154: Please check if we can put these common properties in a variable and
use
```
<TextField
{...commonParams}
>
```

jubitjohn: Okay, I will check

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
```diff
@@ -2,17 +2,49 @@ import SingleInputForm, {
...
 import { CollectionShareSharees } from './sharees';
+import { getLocalCollections } from 'services/collectionService';
 
 export default function EmailShare({ collection }) {
     const galleryContext = useContext(GalleryContext);
 
+    const [updatedOptionsList, setUpdatedOptionsList] = useState(['']);
+    let updatedList = [];
+    useEffect(() => {
+        const owner_user1: User = getData(LS_KEYS.USER);
```
abhinavkgrd, line 20: so we normally use `camelCasing` with variable name

So lets call this `ownerUser`

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
```diff
@@ -2,17 +2,49 @@ import SingleInputForm, {
...
+        const collection_list = getLocalCollections();
```
abhinavkgrd, line 21: we can move this inside the `getUpdatedOptionsList` function and then we could just await inline ?

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
```diff
@@ -2,17 +2,49 @@ import SingleInputForm, {
...
+        const getUpdatedOptionsList = async () => {
+            const result = await collection_list;
+            const emails = result.flatMap((item) => {
+                const shareeEmails = item.sharees.map((sharee) => sharee.email);
+                if (item.owner.email) {
```
abhinavkgrd, line 26: what is the case where `item.owner.email` is not present ?

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
```diff
@@ -2,17 +2,49 @@ import SingleInputForm, {
...
+                    return shareeEmails;
+                }
+            });
+
+            updatedList = Array.from(new Set(emails));
+
+            const filteredList = updatedList.filter(
+                (email) =>
+                    !collection.sharees
+                        .map((sharees) => sharees.email)
```
abhinavkgrd, line 38: we can move this outside the filter, that way we don't we won't have to map over it everytime

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -88,40 +97,86 @@ export default function SingleInputForm(props: SingleInputFormProps) {
...
-                        InputProps={{
-                            endAdornment: props.fieldType === 'password' && (
-                                <ShowHidePassword
-                                    showPassword={showPassword}
-                                    handleClickShowPassword={
-                                        handleClickShowPassword
-                                    }
-                                    handleMouseDownPassword={
-                                        handleMouseDownPassword
+                    {props.optionsList && props.optionsList.length > 0 ? (
```
abhinavkgrd, line 109: can we just always have the `<Autocomplete> ` wrapper and if we don't pass options it would not render any suggestion dropdown ?

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
jubitjohn: Variable name changes to ownerUser

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
jubitjohn: if item.owner.email is not present, that means, it's a collection in our ownership, In that case, I tried to return the sharees mail only

jubitjohn: so here i guess i have made a mistake, If I return both item.owner.email and sharees when owner.email is present , will sharees return those email with which the owner has shared the collection with?

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
jubitjohn: const ownerUser: User = getData(LS_KEYS.USER);
const collection_list = getLocalCollections(); Moved both inside to the function

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
jubitjohn: Yes, I have moved it to a different variable (shareeEmailsCollection) outside the filter

apps/photos/src/components/SingleInputForm.tsx
abhinavkgrd: Ideally we should not be adding a if else and render two types of components from one component. 

So we should ideally have different component for singleInputForm with auto complete

abhinavkgrd:
@jubitjohn , have we added family members to the suggestion list ?

jubitjohn:
Have added family members also

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
abhinavkgrd: but the code is doing something different no?

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
```diff
@@ -1,19 +1,63 @@
...
     const galleryContext = useContext(GalleryContext);
 
-    const collectionShare: SingleInputFormProps['callback'] = async (
+    const [updatedOptionsList, setUpdatedOptionsList] = useState(['']);
+    let updatedList = [];
+    useEffect(() => {
+        const getUpdatedOptionsList = async () => {
+            const ownerUser: User = getData(LS_KEYS.USER);
+            const collection_list = getLocalCollections();
+            const family_list = getLocalFamilyData();
```
abhinavkgrd, line 24: use `camelCased`  variable names

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -49,6 +52,11 @@ export default function SingleInputForm(props: SingleInputFormProps) {
             (message) => setFieldError('inputValue', message),
             resetForm
         );
+
+        if (props.optionsList && props.optionsList.length > 0) {
```
abhinavkgrd, line 56: do we need this here now ?

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -30,9 +30,12 @@ export interface SingleInputFormProps {
     autoComplete?: string;
     blockButton?: boolean;
     hiddenLabel?: boolean;
+    optionsList?: string[];
 }
 
 export default function SingleInputForm(props: SingleInputFormProps) {
+    const [selectedOptions, setSelectedOptions] = useState([]);
```
abhinavkgrd, line 37: also do we need this

apps/photos/src/components/SingleInputForm.tsx
jubitjohn: Not required, removed

apps/photos/src/components/Collections/CollectionShare/emailShare.tsx
jubitjohn: In the case where item.owner.email is present, I guess it makes more sense if I only return  "return [ item.owner.email];", I have a doubt whether current return [...shareeEmails, item.owner.email] also return the list of emails that the owner has shared the album with(that's something we don't need right).

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -30,6 +30,7 @@ export interface SingleInputFormProps {
     autoComplete?: string;
     blockButton?: boolean;
     hiddenLabel?: boolean;
+    optionsList?: string[];
```
abhinavkgrd, line 33: this is not needed no ?

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -49,6 +50,7 @@ export default function SingleInputForm(props: SingleInputFormProps) {
             (message) => setFieldError('inputValue', message),
             resetForm
         );
+
```
abhinavkgrd, line 53: also this new lines too

apps/photos/src/components/SingleInputForm.tsx
```diff
@@ -122,6 +124,7 @@ export default function SingleInputForm(props: SingleInputFormProps) {
                             ),
                         }}
                     />
+
```
abhinavkgrd, line 127: and this one

abhinavkgrd (approved):

PR #1162, opened 2023-06-05
Image exif aspect ratio

abhinavkgrd:
## Description

- Added logic to extract Image dimension from EXIF data
- Added test for image dimension extraction

## Test Plan

- tested with sample upload 
- tested public collect upload
- tested for video, image and live-photo


apps/photos/src/services/upload/uploader.ts
```diff
@@ -116,12 +116,13 @@ export default async function uploader(
...
             metadata.hasStaticThumbnail = true;
         }
-        let pubMagicMetadata: FilePublicMagicMetadata;
-        if (uploaderName) {
-            pubMagicMetadata = await uploadService.constructPublicMagicMetadata(
-                { uploaderName }
-            );
-        }
+
+        const pubMagicMetadata =
```
ua741, line 120: If both uploadName and h/w are missing, will this return nil ?

abhinavkgrd: So, I was adding them as null. 

Have updated the code to ignore `null`/ `undefined` values from magicMetadata

ua741 (approved):

PR #1185, opened 2023-06-13
Navigation-to-corresponding-album--FileInfo

jubitjohn:
## Description

Nativation to corresponding album when the  user clicks on the album cards on the bottom of file info sidebar

### DIFF
--- a/apps/photos/src/components/Chip.tsx
+++ b/apps/photos/src/components/Chip.tsx
@@ -1,14 +1,9 @@
 import { Button, styled } from '@mui/material';
 import { CSSProperties } from '@mui/material/styles/createTypography';
 
-export const Chip = styled(Button)(({ theme }) => ({
-    ...(theme.typography.small as CSSProperties),
-    padding: '8px 12px',
-    borderRadius: '4px',
-    backgroundColor: theme.colors.fill.faint,
-    fontWeight: 'bold',
-    color: 'white',
-    '&:hover': {
-        color: 'black',
-    },
-}));
+export const Chip = styled((props) => <Button color="secondary" {...props} />)(
+    ({ theme }) => ({
+        ...(theme.typography.small as CSSProperties),
+        padding: '8px',
+    })
+);
--- a/apps/photos/src/components/Chip.tsx
+++ b/apps/photos/src/components/Chip.tsx
@@ -1,10 +1,6 @@
-import { Box, styled } from '@mui/material';
+import { Button, styled } from '@mui/material';
 import { CSSProperties } from '@mui/material/styles/createTypography';
 
-export const Chip = styled(Box)(({ theme }) => ({
+export const Chip = styled(Button)(({ theme }) => ({
     ...(theme.typography.small as CSSProperties),
-    padding: '8px 12px',
-    borderRadius: '4px',
-    backgroundColor: theme.colors.fill.faint,
-    fontWeight: 'bold',
 }));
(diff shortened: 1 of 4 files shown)


abhinavkgrd:
@jubitjohn, can you confirm the chip colour and padding are correct as per  
figma https://www.figma.com/file/SYtMyLBs5SAOkTbfMMzhqt/ente-Visual-Design?type=design&node-id=8119-59512&t=nQxLsSFPN6BGdwSW-4

apps/photos/src/components/Chip.tsx
```diff
@@ -1,10 +1,6 @@
-import { Box, styled } from '@mui/material';
+import { Button, styled } from '@mui/material';
 import { CSSProperties } from '@mui/material/styles/createTypography';
 
-export const Chip = styled(Box)(({ theme }) => ({
+export const Chip = styled(Button)(({ theme }) => ({
```
abhinavkgrd, line 4: @jubitjohn, can you confirm the chip colour and padding are correct as per  
figma https://www.figma.com/file/SYtMyLBs5SAOkTbfMMzhqt/ente-Visual-Design?type=design&node-id=8119-59512&t=nQxLsSFPN6BGdwSW-4

apps/photos/src/components/PhotoViewer/FileInfo/index.tsx
```diff
@@ -264,13 +268,21 @@ export function FileInfo({
                                     collectionNameMap.has(collectionID)
                                 )
                                 ?.map((collectionID) => (
-                                    <Chip key={collectionID}>
+                                    <Chip
+                                        key={collectionID}
+                                        onClick={() => {
```
abhinavkgrd, line 273: lets add a helper function `onCollectionChipClick`

apps/photos/src/pages/gallery/index.tsx
```diff
@@ -344,7 +344,7 @@ export default function Gallery() {
         }
         const href = `/gallery${collectionURL}`;
         router.push(href, undefined, { shallow: true });
-    }, [activeCollection]);
+    }, []);
```
abhinavkgrd, line 347: why this ?

jubitjohn: Error: Cancel rendering route: for this Error, Tried removing activeCollection, error was handled , It doesn't seem have affected any other navigations also

abhinavkgrd: But this `useEffect` won't work without the `activeCollection` dependency no? We need to update the navbar URL when the `collectionID` changes. Can you check if there is an option to check if router render is in progress and await on that if possible and then update the route

jubitjohn: setTimeout(() => {
    const href = `/gallery${collectionURL}`;
    router.push(href, undefined, { shallow: true });
  }, 0); Will setting such a timeout work?

abhinavkgrd: How will this help. How have you seen some issues where people have suggested this?

jubitjohn: When setting a timeout, the issue seems to have resolved, Not sure of it's side effects

apps/photos/src/components/Chip.tsx
jubitjohn: Okay, I have checked and made the corrections

apps/photos/src/components/PhotoViewer/FileInfo/index.tsx
jubitjohn: I have added a function

abhinavkgrd (approved):

PR #1189, opened 2023-06-15
Avatar-Component

jubitjohn:
## Description

Avatar icon Created and color scheme set for users based on their UserId


apps/photos/src/components/PhotoFrame.tsx
```diff
@@ -82,6 +103,148 @@ const PhotoFrame = ({
     const [isSourceLoaded, setIsSourceLoaded] = useState(false);
     const [conversionFailed, setConversionFailed] = useState(false);
 
+    const avatarColors: string[] = [
```
ua741, line 106: These colors need to be defined along side with other colors inside theme/ folder.

apps/photos/src/components/PhotoFrame.tsx
```diff
@@ -82,6 +103,148 @@ const PhotoFrame = ({
...
+        '#E99AAD',
+        '#D18484',
+        '#78B5A7',
+    ];
+
+    useEffect(() => {
+        const fetchOwnerEmails = async () => {
+            try {
+                const collection: CollectionItem[] =
+                    await getLocalCollections();
```
ua741, line 136: We can define a method inside collectionSerive that will return userID to email Map. We need the id to email information for sharees as well. In a shared collection, we can have multiple people who might have added photos in a colleciton.

apps/photos/src/components/PhotoFrame.tsx
```diff
@@ -82,6 +103,148 @@ const PhotoFrame = ({
...
+                console.error('Error fetching data:', error);
+            }
+        };
+        filesOfSharedAlbums();
+    }, [sharedAlbums]);
+
+    useEffect(() => {
+        const fileIdToName = async () => {
+            try {
+                const resLocalFiles = getLocalFiles();
```
ua741, line 209: This will result in performance issues when we have large number of files.
We can just pass information about userID to email to the PreviewCard.

Inside the previous card. if file's ownerID is not same as collection ownerID and file's ownerID is present inside the map of userId to email list, we can show the avatar.

apps/photos/src/components/pages/gallery/PreviewCard.tsx
```diff
@@ -323,6 +337,54 @@ export default function PreviewCard(props: IProps) {
         }
     };
 
+    // const [colorCode, setColorCode] = useState('');
+    // const [userLetter, setUserLetter] = useState('');
+
+    const user: User = getData(LS_KEYS.USER);
+
+    // useEffect(() => {
```
ua741, line 345: Remove?

apps/photos/src/services/collectionService.ts
```diff
@@ -1260,3 +1260,31 @@ export async function unhideToCollection(
...
+                sharees.map((item) => {
+                    if (item.id !== user.id)
+                        emailMapping.set(item.id, item.email);
+                });
+            }
+        });
+        const userIdToEmail = emailMapping;
+        return userIdToEmail;
+    } catch (error) {
+        console.error('Error Mapping UserId to email:', error);
```
ua741, line 1288: Use logger for error. Also return empty map?

jubitjohn: added logger, Also added the return type.

apps/photos/src/services/collectionService.ts
```diff
@@ -1260,3 +1260,31 @@ export async function unhideToCollection(
...
+                    if (item.id !== user.id)
+                        emailMapping.set(item.id, item.email);
+                });
+            }
+        });
+        const userIdToEmail = emailMapping;
+        return userIdToEmail;
+    } catch (e) {
+        logError('Error Mapping UserId to email:', e);
+        throw e;
```
ua741, line 1288: To avoid breaking UI, let's return emptyMap.
Or we can define const emailMapping = new Map<number, string>(); before try catch and just return that.

Add a comment saying to avoid breaking UI, we are swallowing the error after logging.

ua741 (approved):

jubitjohn:
Avatar icon Created and color scheme set for users based on their UserId

PR #1212, opened 2023-06-30
Fix slow uploads

abhinavkgrd:
## Description

- Renabled cf worker for uploading files through cf worker proxy similar to download 

## Test Plan

- [x] tested locally 
- [x] pending rate limiting test

### DIFF
--- a/apps/photos/src/services/userService.ts
+++ b/apps/photos/src/services/userService.ts
@@ -19,6 +19,7 @@ import {
     UserDetails,
     DeleteChallengeResponse,
     GetRemoteStoreValueResponse,
+    GetFeatureFlagResponse,
 } from 'types/user';
 import { ServerErrorCodes } from 'utils/error';
 import isElectron from 'is-electron';
@@ -493,3 +494,15 @@ export const updateMapEnabledStatus = async (newStatus: boolean) => {
         throw e;
     }
 };
+
+export async function getDisableCFUploadProxyFlag() {
+    try {
+        const featureFlags = (
+            await fetch('https://static.ente.io/feature_flags.json')
+        ).json() as GetFeatureFlagResponse;
+        return featureFlags.disableCFUploadProxy;
+    } catch (e) {
+        logError(e, 'failed to get feature flags');
+        return undefined;
+    }
+}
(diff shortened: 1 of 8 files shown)


apps/photos/src/services/userService.ts
```diff
@@ -493,3 +494,15 @@ export const updateMapEnabledStatus = async (newStatus: boolean) => {
...
+
+export async function getDisableCFUploadProxyFlag() {
+    try {
+        const featureFlags = (
+            await fetch('https://static.ente.io/feature_flags.json')
+        ).json() as GetFeatureFlagResponse;
+        return featureFlags.disableCFUploadProxy;
+    } catch (e) {
+        logError(e, 'failed to get feature flags');
+        return undefined;
```
ua741, line 506: Why not return a default value here with CF worked as disabled or enabled

apps/photos/src/services/userService.ts
```diff
@@ -493,3 +494,15 @@ export const updateMapEnabledStatus = async (newStatus: boolean) => {
         throw e;
     }
 };
+
+export async function getDisableCFUploadProxyFlag() {
```
ua741, line 498: Missing return type

apps/photos/src/utils/common/apiUtil.ts
```diff
@@ -45,7 +45,7 @@ export const getPublicCollectionThumbnailURL = (id: number) => {
 };
 
 export const getUploadEndpoint = () => {
-    const endpoint = process.env.NEXT_PUBLIC_ENTE_ENDPOINT;
+    const endpoint = process.env.NEXT_PUBLIC_ENTE_UPLOAD_ENDPOINT;
```
ua741, line 48: What's this url change? As someone  without much understanding of the code base, I think based on flag we would choose different values

apps/photos/src/services/userService.ts
abhinavkgrd: ahh yes will update

apps/photos/src/utils/common/apiUtil.ts
abhinavkgrd: so, I updated the add a new variable -> `NEXT_PUBLIC_ENTE_UPLOAD_ENDPOINT`, earlier it was api was never used and was pointing to the ENTE_ENDPOINT as default value.

ua741 (approved):

PR #1221, opened 2023-07-03
Add to collab album support

abhinavkgrd:
## Description
- Added support to add to collab album 

## Test Plan

tested locally

### DIFF
--- a/apps/photos/src/services/collectionService.ts
+++ b/apps/photos/src/services/collectionService.ts
@@ -48,12 +48,13 @@ import { User } from 'types/user';
 import {
     isQuickLinkCollection,
     isOutgoingShare,
-    isIncomingShare,
     isSharedOnlyViaLink,
     isValidMoveTarget,
     isHiddenCollection,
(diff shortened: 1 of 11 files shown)


apps/photos/src/services/collectionService.ts
```diff
@@ -1024,9 +1025,11 @@ export async function getCollectionSummaries(
                 latestFile: collectionLatestFiles.get(collection.id),
                 fileCount: collectionFilesCount.get(collection.id) ?? 0,
                 updationTime: collection.updationTime,
-                type: isIncomingShare(collection, user)
-                    ? CollectionSummaryType.incomingShare
-                    : isOutgoingShare(collection)
+                type: isIncomingViewerShare(collection, user)
+                    ? CollectionSummaryType.incomingShareViewer
```
ua741, line 1029: Can we handle for newer roles? Ideally, if the RoleType is Collaborator, use incomingShareCollaborator otherwise default to Viewer

ua741 (approved):

PR #1224, opened 2023-07-04
Show incoming shared files in ALL section 

abhinavkgrd:
## Description
- added support to show incoming shared files in all section
- added support to search through shared files

## Test Plan

- tested that incoming shared files are showed in the all section
- selection works properly
- in PhotoVeiwer view has correct options


ua741 (commented):
Have we also tested it for incoming collection that are archived?

abhinavkgrd:
> Have we also tested it for incoming collection that are archived?

Have created separate PR to handle this https://github.com/ente-io/photos-web/pull/1236

PR #1238, opened 2023-07-10
Album cover

abhinavkgrd:
## Description

- Added support to show album covers, if the user has set them
- else show the first file based on the sort order 

## Test Plan

tested locally

### DIFF
--- a/apps/photos/src/services/collectionService.ts
+++ b/apps/photos/src/services/collectionService.ts
@@ -277,27 +277,38 @@ export const getCollection = async (
     }
 };
 
-export const getCollectionLatestFiles = (
+export const getCollectionCoverFiles = (
     user: User,
     files: EnteFile[],
-    archivedCollections: Set<number>
+    archivedCollections: Set<number>,
+    collections: Collection[]
 ): CollectionLatestFiles => {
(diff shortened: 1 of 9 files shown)


apps/photos/src/services/collectionService.ts
```diff
@@ -300,6 +300,40 @@ export const getCollectionLatestFiles = (
...
+        collectionMap.set(collection.id, collection);
+    });
+
+    files.forEach((file) => {
+        const collection = collectionMap.get(file.collectionID);
+        if (collection.pubMagicMetadata?.data?.coverID === file.id) {
+            coverFiles.set(file.collectionID, file);
+        } else if (collection.pubMagicMetadata?.data?.asc) {
+            coverFiles.set(file.collectionID, file);
+        } else if (!coverFiles.has(file.collectionID)) {
```
ua741, line 322: If first file is cover and next is not, the else if logic will override the cover?

abhinavkgrd: It should not. 
 It checks if `coverFile` is `not` set.

ua741 (approved):

PR #1301, opened 2023-08-07
Add delete option to search mode

abhinavkgrd:
## Description

^  title 

## Test Plan

tested locally


ua741 (approved):
When there are mixed set of files (with different owners), what will be the behavior? We will only trash files owned by the current user?

abhinavkgrd:
> When there are mixed set of files (with different owners), what will be the behavior? We will only trash files owned by the current user?

It has the same behavior as selecting and deleting from all section.

PR #1377, opened 2023-09-26
reposition convert button

abhinavkgrd:
## Description

Moved the convert button to the top action bar

image

## Test Plan

tested locally


ua741:
@vishnukvmd This will reduce the discoverability of this option. It might be similar situation as current map option on mobile. cc @ashilkn

vishnukvmd:
Ideally we should have plugged this into the video control bar (next to the play button), but since that's more work, this seemed like an okay starting point

ua741:
> Ideally we should have plugged this into the video control bar (next to the play button), but since that's more work, this seemed like an okay starting point

Hmm. I don't think we would spend time on moving this to video control bar in the near future.
Personally, don't think it's a good idea but can go ahead with it.

ua741 (approved):

PR #1402, opened 2023-10-25
Add logs for gif upload issue debugging

abhinavkgrd:
## Description

^ title 

## Test Plan

tested locally

### DIFF
--- a/apps/photos/src/services/HTTPService.ts
+++ b/apps/photos/src/services/HTTPService.ts
@@ -1,4 +1,6 @@
 import axios, { AxiosRequestConfig } from 'axios';
+import { addLogLine } from 'utils/logging';
+import { logError } from 'utils/sentry';
 
 interface IHTTPHeaders {
     [headerKey: string]: any;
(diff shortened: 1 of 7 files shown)


apps/photos/src/services/HTTPService.ts
```diff
@@ -15,12 +17,38 @@ class HTTPService {
...
         axios.interceptors.response.use(
             (response) => Promise.resolve(response),
-            (err) => {
-                if (!err.response) {
-                    return Promise.reject(err);
+            (error) => {
+                if (error.response) {
+                    // The request was made and the server responded with a status code
+                    // that falls out of the range of 2xx
+                    logError(error, 'HTTP Service Error', {
```
ua741, line 24: Will this log on sentry? If yes, do we really want it for all requests?

abhinavkgrd: yes, it will be logged on sentry and No harm in having them no?...will help with debugging?

ua741: Hmm, I feel it would increase the noise, but could be wrong as well. You would know better.

ua741 (approved):

PR #1494, opened 2023-12-07
Add File disk caching

abhinavkgrd:


apps/photos/src/services/download/index.ts
```diff
@@ -123,6 +128,34 @@ class DownloadManager {
             throw e;
         }
     }
+    private async getCachedFile(file: EnteFile): Promise<Response> {
+        try {
+            const cacheResp: Response = await this.fileCache?.match(
+                file.id.toString()
+            );
+            // check if cached file size is same as file size
+            if (cacheResp) {
```
ua741, line 137: early return

apps/photos/src/services/download/index.ts
```diff
@@ -123,6 +128,34 @@ class DownloadManager {
             throw e;
         }
     }
+    private async getCachedFile(file: EnteFile): Promise<Response> {
+        try {
+            const cacheResp: Response = await this.fileCache?.match(
```
ua741, line 133: early terminate

apps/photos/src/services/download/index.ts
```diff
@@ -85,6 +89,7 @@ class DownloadManager {
             }
             this.downloadClient = createDownloadClient(app, tokens, timeout);
             this.thumbnailCache = await openThumbnailCache();
+            this.fileCache = await openFileCache();
```
ua741, line 92: `diskFileCache`. Mark nullable and comment that only available in electron

apps/photos/src/services/download/index.ts
```diff
@@ -494,6 +543,15 @@ async function openThumbnailCache() {
     }
 }
 
+async function openFileCache() {
+    try {
+        return await CacheStorageService.open(CACHES.FILES, FILE_CACHE_LIMIT);
+    } catch (e) {
+        logError(e, 'Failed to open file cache');
+        return null;
```
ua741, line 551: Swallow only for external users. Ideally we should know internally when it's failing

ua741 (approved):

PR #1523, opened 2023-12-28
Accounts-passkeys

httpjamesm:
## Description

* Adds Ente Accounts
* Adds passkey enrolment, renaming and deleting


apps/accounts/public/fonts/inter-v11-latin-500.woff
ua741, line 1: Do we need to add these fonts?

httpjamesm: Would you rather use next/font? Vishnu didn't like the default font look because it wasn't brand consistent

PR #1543, opened 2024-01-10
Onnx clip UI

abhinavkgrd:
## Description
- Added logic to conditionally load and sync  embeddings, 
- Updated clip service to support `onnx-clip` model 
- Migrated local embeddings table to remove drop old `ggml-clip` embeddings

## Test Plan

tested locally


apps/photos/src/services/clipService.ts
```diff
@@ -166,7 +170,7 @@ class ClipServiceImpl {
                 return;
             }
             const localFiles = getPersonalFiles(await getAllLocalFiles(), user);
-            const existingEmbeddings = await getLatestClipImageEmbeddings();
+            const existingEmbeddings = await getLocalEmbeddings(model);
```
abhinavkgrd, line 173: Have updated to `getLocalEmbeddings`, because we anyways are blocking the scheduling of clip embedding extraction on remote sync to finish. So, local embedding will be the latest already.

apps/photos/src/services/embeddingService.ts
```diff
@@ -21,26 +22,31 @@ const ENDPOINT = getEndpoint();
...
+    if (!embeddings) {
+        await localForage.removeItem(EMBEDDINGS_TABLE_V1);
+        await localForage.setItem(EMBEDDINGS_TABLE, []);
+        await localForage.setItem(EMBEDDING_SYNC_TIME_TABLE, 0);
+        return [];
+    }
+    if (model) {
+        return embeddings.filter((embedding) => embedding.model === model);
+    } else {
+        return embeddings;
```
0nullpointer, line 41: Why do we return embeddings if the model is empty?

abhinavkgrd: so, we needed `allEmbeddings` at some places like, `syncEmbedding` and `cleanupDeletedEmbeddings`, that's why I let not passing any model be the indicator that the caller wants all the embeddings.

Have updated the to have separate `getAllEmbeddings`, as that is safer

0nullpointer (approved):

PR #1548, opened 2024-01-11
fix missing await

abhinavkgrd:
## Description

^ title 

## Test Plan

tested locally


vishnukvmd (approved):

vishnukvmd:
If we can add a lint rule for missing awaits, we should. It's been helpful on mobile.

PR #1560, opened 2024-01-18
Display full name of file and collection on hover

### LINKED ISSUE #1559: Display complete photo name and Album on hover in deduplicate
It would be nice to easily check the photo location and name to choose if it should be removed or not. I think that showing this information on hover would make the process easier because it would not interact with the current selection.

abhinavkgrd:
## Description

fixes #1559

^ title 

image

## Test Plan

tested locally

### DIFF
--- a/apps/photos/src/components/pages/gallery/PreviewCard.tsx
+++ b/apps/photos/src/components/pages/gallery/PreviewCard.tsx
@@ -1,6 +1,6 @@
 import React, { useContext, useEffect, useRef, useState } from 'react';
 import { EnteFile } from 'types/file';
-import { styled } from '@mui/material';
+import { Tooltip, styled } from '@mui/material';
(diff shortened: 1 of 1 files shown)


0nullpointer:
@vishnukvmd For very large filenames like below, I am not sure how to format the text on hover since the collection name gets drowned deep down in the text
Screenshot 2024-01-18 at 12 03 43

And maybe we can do an centre align for the collection name if we are going forward with this nonetheless,
Screenshot 2024-01-18 at 12 01 16

vishnukvmd:
@0nullpointer that's alright for now, this is a short-term fix. We should consolidate the implementation across mobile and desktop to add symlinks during deduplication so that users don't have to choose which instance of a file they wish to retain.

0nullpointer (approved):

PR #1599, opened 2024-02-03
Deploy auth from the same repository

mnvr:
Not sure how to test this other than by deploying once.


mnvr:
Symlinks don't seem to be working, so I copied the output directory. I did a test deployment to a test-branch, and that works fine.

abhinavkgrd (approved):

